PR #6, opened 2021-06-17
chore: Cleanup of unnecessary and redundant files. Adding GH actions

scalvert:
As the description suggests.


glnster (approved):
thank you for doing this!

Issue #8, opened 2021-06-17
switch to eslint


glnster:
switch from xo to eslint

PR #9, opened 2021-06-18
remove xo, install eslint, fix lint errors

glnster:
# Summary
Taking a stab at addressing #8 - Remove xo, install eslint, and fix any lint errors.

# Testing Done
`yarn test`


package.json
```diff
@@ -17,34 +17,24 @@
     "prepublishOnly": "npm run build",
     "prebuild": "rimraf dist",
     "build": "babel src --out-dir dist --copy-files --ignore **/*.test.js",
-    "lint": "xo",
+    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
+    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint:js": "eslint . --ext js,ts --cache",
+    "lint:js:fix": "eslint . --ext js,ts --fix",
```
scalvert, line 23: I'd probably just have the `lint:js` script, but just name it lint and dump all the others. Using npm-run-all is usually used if you run multiple linters in a project. We're only running one, so we don't really need all that extra stuff.

package.json
```diff
@@ -69,14 +61,20 @@
     "babel-jest": "^23.6.0",
     "babel-plugin-add-module-exports": "^1.0.0",
     "babel-plugin-tester": "^10.1.0",
+    "eslint": "^7.28.0",
+    "eslint-config-standard": "^16.0.3",
+    "eslint-plugin-import": "^2.23.4",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prettier": "^3.4.0",
```
scalvert, line 68: I think you also want to add `eslint-config-prettier`, and [define prettier as the last plugin in your extends](https://github.com/prettier/eslint-config-prettier#installation).

scalvert (changes requested):

.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  env: {
+    browser: false,
+    node: true,
+  },
+  plugins: ['node', 'prettier'],
+  extends: [
+    'eslint:recommended',
+    'plugin:node/recommended',
+    'plugin:prettier/recommended',
```
scalvert, line 10: I think if you use `eslint-config-prettier`, you just specify `"prettier",` here, and it'll correctly configure it for you.

.eslintrc.js
```diff
@@ -0,0 +1,17 @@
...
+  ],
+  parserOptions: {
+    ecmaVersion: 12,
```
scalvert, line 13: I think this isn't right? I think we want 2019 or above.

__tests__/index.spec.js
```diff
@@ -1,56 +1,136 @@
 const pluginTester = require('babel-plugin-tester').default;
```
scalvert, line 1: Can we rename this file to `index-test.js`. That seems to be a more conventional name.

package.json
```diff
@@ -69,14 +58,21 @@
...
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-config-standard": "^16.0.3",
+    "eslint-plugin-import": "^2.23.4",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prettier": "^3.4.0",
+    "eslint-plugin-promise": "^5.1.0",
     "husky": "^1.2.0",
     "jest": "^23.6.0",
     "lint-staged": "^8.1.0",
+    "npm-run-all": "^4.1.5",
```
scalvert, line 71: Do we need this?

package.json
```diff
@@ -17,11 +17,10 @@
...
     "prebuild": "rimraf dist",
     "build": "babel src --out-dir dist --copy-files --ignore **/*.test.js",
-    "lint": "xo",
+    "lint": "eslint . --ext js,ts --cache",
     "postbuild": "rimraf dist/**/*.test.js",
-    "test:watch": "jest --watch",
-    "test:xo": "xo && jest --coverage",
-    "test": "jest"
+    "test": "npm-run-all lint test:*",
+    "test:jest": "jest"
```
scalvert, line 23: I think we don't need npm-run-all.

```suggestion
    "test": "yarn lint && yarn test:jest",
    "test:jest": "jest"
```

scalvert: You only need `npm-run-all` if you have multiple things you want to run, and specifically use glob-like patterns to run them. Also, it parallelizes things, and I don't think we need that here.

package.json
```diff
@@ -17,11 +17,10 @@
     "prepublishOnly": "npm run build",
     "prebuild": "rimraf dist",
```
scalvert, line 18: Does `rm -rf ...` not work here?

scalvert (changes requested):

.eslintrc.js
glnster: 12 = 2021
From their [docs](https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments):

_ecmaVersion - set to 3, 5 (default), 6, 7, 8, 9, 10, 11, or 12 to specify the version of ECMAScript syntax you want to use. You can also set to 2015 (same as 6), 2016 (same as 7), 2017 (same as 8), 2018 (same as 9), 2019 (same as 10), 2020 (same as 11), or 2021 (same as 12) to use the year-based naming._

scalvert: So this is very confusing to read. I think we should use the year. I've never seen it configured this way before (and people may thing the config is broken).

scalvert (changes requested):

scalvert (approved):
Thanks for making these changes!

PR #14, opened 2021-07-01
Part 1 Fix beforeEachModified conditional checking

glnster:
Refactor to fix issues where `beforeEachModified` conditional checking wasn't working as expected, as in some test cases, it was getting reset prematurely.

Updated behavior such that if the plugin needs to create a new beforeEach, it inserts it either after any test setup calls or at the top of the test module.

As part of the updated behavior, the refactor also includes checking for any custom "hooks" name passed into the test module's function.

Testing done on:
- this stand-alone repo
- a simple ember app
- a test file in a larger, private ember app


src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
...
+              nodeName = babelPath.node.name;
+            }
+
+            isFirstChildTestMethodCall =
+              nodeName &&
+              testMethodCalls.includes(nodeName) &&
+              babelPath.scope.path.parentPath &&
+              babelPath.scope.path.parentPath.node &&
+              babelPath.scope.path.parentPath.node.callee &&
+              babelPath.scope.path.parentPath.node.callee.name === 'module';
```
ckundo, line 222: This drilling into properties feels like a smell to me. I'm not sure yet what approach would make this more succinct and DRY..

ckundo:
Are there unit tests for this project. That might help catch these kinds of cases.

ckundo (approved):
would like to see a unit test if there's already infrastructure set up in the project for that. If not maybe a follow on to add unit testing?

glnster:
> Are there unit tests for this project. That might help catch these kinds of cases.

There are tests, but they are snapshot-based and not as strict unit tests. I'm using https://github.com/babel-utils/babel-plugin-tester as the test infra. As cases come up, I add a test input and verify the snapshot output. Offline and outside of this project I also test on a simple ember app.

__tests__/__fixtures__/multiple-module-no-beforeeach-test-skip-todo-input-test.js
```diff
@@ -19,14 +19,14 @@ module('Acceptance | browse acceptance test', function (hooks) {
...
-module('Acceptance | browse acceptance test', function (hooks) {
+module('Acceptance | search acceptance test', function (hooks) {
   setupApplicationTest(hooks);
 
   some.otherThing(function () {
     noop();
   })
 
-  test.skip('it renders browse page', async function (assert) {
+  test.skip('it renders search page', async function (assert) {
```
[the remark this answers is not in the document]
glnster: The "skip" doesn't have any affect as all of these tests are fixtures for snapshots. They are only meant to represent code that the plugin will be transforming and they don't actually execute.
https://github.com/babel-utils/babel-plugin-tester

__tests__/__fixtures__/multiple-module-no-beforeeach-test-skip-todo-input-test.js
```diff
@@ -19,14 +19,14 @@ module('Acceptance | browse acceptance test', function (hooks) {
   });
 });
 
-module('Acceptance | browse acceptance test', function (hooks) {
+module('Acceptance | search acceptance test', function (hooks) {
   setupApplicationTest(hooks);
 
   some.otherThing(function () {
     noop();
```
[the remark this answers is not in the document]
scalvert: No, it's sync, and also not material to the test case that's being exercised.

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
```diff
@@ -0,0 +1,25 @@
+import {module, test} from 'qunit';
+import {setupApplicationTest} from 'ember-qunit';
+
+const SELECTORS = Object.freeze({
+  MOCK_SELECTOR: '[data-test-nav-bar-browse]',
+});
+
+module('Acceptance | browse acceptance test', function (hooks) {
```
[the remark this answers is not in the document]
scalvert: I'm not sure this matters. These are test fixtures, and never really intended to be read (and are also in the __fixtures__ dir).

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
```diff
@@ -0,0 +1,25 @@
...
+  setupApplicationTest(hooks);
+
+  some.otherThing(function () {
```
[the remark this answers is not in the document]
scalvert: Correct. We need to make sure we're transpiling correctly, irrespective of other code in the test module.

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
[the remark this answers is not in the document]
glnster: Thanks for the comment. Are you referring to the fixture's file name? If so, I will keep the file name as is for now, to be consistent with the rest of the test fixture names. Eventually I'd like to break apart these test fixtures if possible so that they are representing and testing only a specific function of the plugin transformation. I'd also like to make these file name references more dynamic.

__tests__/__fixtures__/multiple-module-no-beforeeach-test-skip-todo-input-test.js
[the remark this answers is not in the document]
glnster: No, these tests are only mocks / fixtures and don't execute.

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
```diff
@@ -0,0 +1,25 @@
...
+    noop();
+  })
+
+  hooks.beforeEach(async function () {
+    const myConst = 0;
+    await noop();
+    // do some things here
```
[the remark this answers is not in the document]
scalvert: This is again more about positional transpilation. I believe @glnster is trying to verify the position of his injected code. Correct me if I'm wrong here, Glenn.

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  test('it renders browse page', async function (assert) {
```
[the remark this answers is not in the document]
scalvert: I'm not sure what you mean here. These tests never run, as they're not invoked via QUnit or the Ember test command. These fixtures are 'inert'.

src/index.js
```diff
@@ -92,7 +92,11 @@ function shouldLoadFile(filename) {
 }
 
 function hasMetadataDeclaration({ node }) {
-  if (node.expression && node.expression.type === 'AssignmentExpression') {
+  if (
+    node.expression &&
+    node.expression.type === 'AssignmentExpression' &&
```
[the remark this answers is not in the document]
scalvert: There's an `isAssignmentExpression` on babel types.

src/index.js
```diff
@@ -124,6 +128,8 @@ function addMetadata({ types: t }) {
           return;
         }
 
+        state.opts.transformedModules = [];
```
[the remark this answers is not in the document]
scalvert: This is the idiomatic way to include stateful data when doing recursive tree traversals. We need a way to provide data to the subsequent visitor, and this provides that mechanism.

src/index.js
scalvert: The only way to do this would to use a helper that allows for inspecting a path. 

Example:

```js
function get(node, path) {
  if (!node) {
    return;
  }

  let parts;
  if (typeof path === 'string') {
    parts = path.split('.');
  } else {
    parts = path;
  }

  if (parts.length === 1) {
    return node[path];
  }

  let property = node[parts[0]];

  if (property && parts.length > 1) {
    parts.shift();
    return get(property, parts);
  }

  return property;
}
```

Which you'd call like so:

```js
get(babelPath, 'scope.path.parentPath.node.callee.name')
```

src/index.js
```diff
@@ -92,7 +92,11 @@ function shouldLoadFile(filename) {
 }
 
 function hasMetadataDeclaration({ node }) {
-  if (node.expression && node.expression.type === 'AssignmentExpression') {
```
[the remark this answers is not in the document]
scalvert: Good suggestion. Should we also not be checking for whether `node.expression.left.property` is truthy?

src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
       CallExpression(babelPath, state) {
         if (!state.opts.shouldLoadFile) return;
 
-        // If we're at a top-level call expression, then we reset the beforeEachModified state to false, and let Babel
-        // move on to the next call expression.
-        if (babelPath.scope.block.type === 'Program') {
-          state.opts.beforeEachModified = false;
-          return;
+        // If this call expression is a top-level module, store string arg, else skip the nested module entirely
```
scalvert, line 185: I think you mean 'store module name string' rather than 'store string arg'.

```suggestion
        // If this call expression is a top-level module, store module name string, else skip the nested module entirely
```

src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
...
+        if (babelPath.get('callee').isIdentifier({ name: 'module' })) {
+          if (babelPath.parentPath.parent.type === 'Program') {
```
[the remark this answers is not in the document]
scalvert: I realize this makes it a bit more terse, but I find it makes this harder to read. I now have to do the conditional in my head rather than just reading the cascade. 

In most cases I'd agree with you, in this case I don't.

src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
...
 
-        if (!state.opts.beforeEachModified) {
-          const hasBeforeEach = babelPath.node.callee.property
+        if (
+          state.opts.moduleName &&
+          !state.opts.transformedModules.includes(state.opts.moduleName)
+        ) {
+          const isBeforeEach = babelPath.node.callee.property
             ? babelPath.node.callee.property.name === 'beforeEach'
             : false;
```
scalvert, line 201: It may be nice to split these types of checks out in to separate helpers, so we lessen the logic within each visitor.

src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
...
-          let nodeName = '';
-
-          if (babelPath.node.callee.name) {
-            nodeName = babelPath.node.callee.name;
-          } else if (babelPath.node.callee.object) {
-            nodeName = babelPath.node.callee.object.name;
+          let isFirstChildTestMethodCall;
+
+          if (!isBeforeEach) {
+            const testMethodCalls = ['test', 'tests', 'module'];
```
scalvert, line 205: I'd probably move this out into a module scoped const, unless of course you're mutating this somewhere else (which I can't seem to see).

src/index.js
```diff
@@ -176,37 +182,49 @@ function addMetadata({ types: t }) {
...
+            let nodeName = '';
+
+            if (babelPath.node.callee && babelPath.node.callee.name) {
+              nodeName = babelPath.node.callee.name;
+            } else if (babelPath.node.callee.object) {
+              nodeName = babelPath.node.callee.object.name;
+            } else {
+              nodeName = babelPath.node.name;
+            }
```
scalvert, line 214: This seems ripe for extracting to a helper.

scalvert (changes requested):

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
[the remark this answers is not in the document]
ckundo: thanks, that answers my question.. wanted to make sure we weren't running everything in `__tests__` with a glob pattern for `*.js`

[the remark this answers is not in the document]
scalvert: Yep makes sense. We'd need to be using the same framework that is invoking the repo's tests, which is `jest`, in order to run these.

__tests__/__fixtures__/one-module-one-async-beforeeach-new-import-input-test.js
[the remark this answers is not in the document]
glnster: Yes, that's correct. It is just a mock comment to test how the plugin handles those lines and where it positions the statements it inserts.

src/index.js
```diff
@@ -1,5 +1,7 @@
 const path = require('path');
 
+const TEST_METHOD_CALL_NAMES = ['test', 'tests', 'module'];
```
scalvert, line 3: Where does the `tests` API come from?

src/index.js
```diff
@@ -91,15 +93,73 @@ function shouldLoadFile(filename) {
...
     return false;
   }
+
+  return (
+    node.expression.left.object.name === 'testMetadata' &&
+    node.expression.left.property.name === 'filePath'
+  );
+}
+
+function getNodeProperty(node, path) {
```
scalvert, line 112: I'd probably have this in a separate util with some quick tests to sanity check.

src/index.js
```diff
@@ -91,15 +93,73 @@ function shouldLoadFile(filename) {
...
+    return callee.name;
+  } else if (callee.object) {
+    return callee.object.name;
+  } else {
+    return;
+  }
+}
+
+function isBeforeEach(callee) {
+  return callee.property ? callee.property.name === 'beforeEach' : false;
```
scalvert, line 149:
```suggestion
  return callee.property && callee.property.name === 'beforeEach';
```
?

src/index.js
```diff
@@ -176,37 +238,35 @@ function addMetadata({ types: t }) {
...
-            babelPath.scope.path.parentPath.node.callee.name === 'module';
-
-          const shouldDoTransform = hasBeforeEach || isFirstChildTestMethodCall;
+        if (
+          state.opts.moduleName &&
+          !state.opts.transformedModules.includes(state.opts.moduleName)
+        ) {
+          const callee = babelPath.node.callee;
+          const calleeName = getCalleeName(callee);
+          const nodeName = calleeName ? calleeName : babelPath.node.name;
```
scalvert, line 257:
```suggestion
          const nodeName = calleeName || babelPath.node.name;
```
?

scalvert (changes requested):

src/index.js
glnster: Discussed offline. Thanks Steve. Instead of searching against these hard coded call names, I'll take the approach we discussed.

ckundo (approved):
Curious about the file matching regex and the potential edge cases with the `getLastSetupCall`. If those are not relevant, LGTM.

__tests__/__fixtures__/one-module-no-beforeeach-no-setup-test.js
```diff
@@ -0,0 +1,14 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { click, visit } from '@ember/test-helpers';
+
+const SELECTORS = Object.freeze({
+  MOCK_SELECTOR: '[data-test-nav-bar-browse]',
+});
+
+module('Acceptance | browse acceptance test', function (hooks) {
```
scalvert, line 9: This shouldn't have a hooks param either, right?

__tests__/__fixtures__/one-module-one-beforeeach-no-hooks-new-import-input-test.js
```diff
@@ -0,0 +1,11 @@
+import { module, test } from 'qunit';
```
scalvert, line 1: This test is called `one-module-one-beforeeach-no-hooks-new-import-input-test.js`, but it doesn't appear to have one beforeEach. Was that implying the begin state, or the end state?

__tests__/utils-test.js
```diff
@@ -0,0 +1,24 @@
...
+    };
+
+    expect(getNodeProperty(mockNode, 'system.settings.volume.level')).toBe(11);
+    expect(getNodeProperty(mockNode.system.settings, 'volume.level')).toBe(11);
+    expect(getNodeProperty(mockNode, 'unknown.leaf.property')).toBe(undefined);
+  });
+
+  it('returns early if no node is provided', () => {
+    let emptyNode;
+    expect(getNodeProperty(emptyNode, 'volume.level')).toBe(undefined);
```
scalvert, line 22: I'd probably add:

```suggestion
    expect(getNodeProperty(emptyNode, 'volume.level')).toBe(undefined);
    expect(getNodeProperty(null, 'volume.level')).toBe(undefined);
    expect(getNodeProperty({}, 'volume.level')).toBe(undefined);
```

src/index.js
```diff
@@ -1,51 +1,132 @@
...
+    return callsArray[0];
 
-    if (functionBlockBodyStatementsArray.length > 0) {
-      existingMetadataDeclaration = functionBlockBodyStatementsArray.find(
-        hasMetadataDeclaration
-      );
+  for (let i = 0; i < callsArray.length; i++) {
+    if (
+      callsArray[i + 1] &&
+      !isSetupCall(callsArray[i + 1], t, hooksIdentifier)
```
scalvert, line 35: As a rule of thumb (my thumbs!), I always store array access in a variable if I'm accessing it more than once. It also adds to readability, but for sure removes instances of bugs happening when I change one indexers value.

src/index.js
```diff
@@ -1,51 +1,132 @@
...
-      null,
-      [],
-      t.blockStatement([testMetadataVarDeclaration, testMetadataAssignment])
-    );
-    const beforeEachExpression = t.expressionStatement(
-      t.callExpression(
-        t.memberExpression(t.identifier('hooks'), t.identifier('beforeEach')),
-        [beforeEachFunc]
-      )
+  }
```
scalvert, line 61: I'd probably use a `for..of` for this 🤷‍♂️

src/index.js
```diff
@@ -1,51 +1,132 @@
...
 
+/**
+ * Creates a new beforeEach call with the test metadata statements.
+ * Inserts this new beforeEach below any existing setup calls, else
+ * at the top of the test module function block.
+ * Updates state.opts.transformedModules [] with the name of the test module transformed.
+ * @param {object} state - Babel state
+ * @param {object} t  - Babel types
+ */
+function writeNewBeforeEach(state, t) {
```
scalvert, line 102: `insertBeforeEach` feels more idiomatic to me in terms of naming for babel type stuff. WDYT?

src/index.js
```diff
@@ -87,19 +168,35 @@ function getTestMetadataDeclaration(state, t) {
...
+ */
+function hasMetadataDeclaration({ node }, t) {
+  if (
+    !node.expression ||
+    !t.isAssignmentExpression(node.expression) ||
+    !node.expression.left.object ||
+    !node.expression.left.property
+  ) {
     return false;
   }
```
scalvert, line 185: I may be inclined to store this in a variable, and && it off the boolean below, in which case it'd short circuit if `false`.

src/index.js
```diff
@@ -157,13 +260,14 @@ function addMetadata({ types: t }) {
...
-       *   2. if it's a call to either "test" or to a nested "module", then we want to add a new beforeEach just above it
-       * If the current call is neither of these, then do nothing. Babel will move on to the next call expression.
+       * While inside a top-level module, we:
+       * 1. check for any existing beforeEach call
+       * 2. else, check for any test setup call expressions, e.g. setupApplication(hooks)
+       *
+       * If there is an existing beforeEach, we add our statements to it. Otherwise, we add a new beforeEach, either
+       * after any test setup calls or at the top of the test module.
        *
        * The transformed beforeEach would look like:
```
scalvert, line 272: I worry this comment is going to age out, as evidenced by your having to keep updating it. I'd much prefer to keep the code with semantically named variables so they express the intent.

scalvert (changes requested):

src/index.js
```diff
@@ -1,51 +1,132 @@
...
+  if (callsArray.length === 1 && isSetupCall(callsArray[0], t, hooksIdentifier))
+    return callsArray[0];
 
-    if (functionBlockBodyStatementsArray.length > 0) {
-      existingMetadataDeclaration = functionBlockBodyStatementsArray.find(
-        hasMetadataDeclaration
-      );
+  for (let i = 0; i < callsArray.length; i++) {
+    if (
+      callsArray[i + 1] &&
```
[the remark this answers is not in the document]
glnster: I think I'd have to see more of the surrounding stuff to get more context. But if it was something like:
`module('Integration | Component | bar', function(hooks) { 
  setupRenderingTest(hooks);
  setupTwo(hooks);
  setupThree(hooks);

  hooks.beforeEach(function(){...});
  hooks.afterEach();
  hooks.beforeEach();
`
The plugin, traversing top down on each call, would check each of the `anything(hooks)` calls and store the last one, setupThree(hooks).

Then it would check for any beforeEach calls and store the first beforeEach. That's the beforeEach it would add our statements to.

src/index.js
```diff
@@ -87,19 +168,35 @@ function getTestMetadataDeclaration(state, t) {
...
+  );
+}
+
+/**
+ * Checks for files ending with "-test.js" or "_test.js"
+ * @param {string} filename File name which may include file path
+ * @returns {Boolean}
+ */
+function shouldLoadFile(filename) {
+  return filename.match(/[-_]test\.js/gi);
```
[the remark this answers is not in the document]
glnster: There's bound to be this case. I'll need to work on this regex.

__tests__/__fixtures__/one-module-one-beforeeach-no-hooks-new-import-input-test.js
glnster: Ah, copy typo. The file names should imply the begin state.

src/index.js
glnster: I agree. thank you

src/index.js
glnster: Ok, I cleaned it up and only left the part about sibling/nested modules and how the beforeEach would look like.

src/index.js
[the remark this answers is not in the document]
glnster: I think this can be follow up work. `/[-_](test|spec)\.js/gi`

package.json
```diff
@@ -63,7 +63,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.4.0",
     "husky": "^1.2.0",
-    "jest": "^23.6.0",
+    "jest": "^27.0.6",
```
scalvert, line 66: Any reason why you downgraded?

src/index.js
```diff
@@ -1,51 +1,134 @@
...
+}
+
+/**
+ * Adds test metadata statements to the top of an existing beforeEach call function block.
+ * Updates state.opts.transformedModules [] with the name of the test module transformed.
+ * @param {object} state - Babel state
+ * @param {object} beforeEachExpression - the beforeEach Babel call expression
+ * @param {object} t Babel types
+ */
+function addMetaDataToBeforeEach(state, beforeEachExpression, t) {
```
scalvert, line 73:
```suggestion
function insertMetaDataInBeforeEach(state, beforeEachExpression, t) {
```

src/index.js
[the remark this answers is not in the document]
scalvert: Spec isn't something that's used in an ember project. Generally, tests are either initially scaffolded via ember-cli, which creates tests in a `tests` directory, or added via the generators, which adds them to the `tests` directory. I've never once seen anyone use `spec` in an ember project, which is what we're targeting.

scalvert (approved):

scalvert:
Just a few small changes, but overall it looks GTG! Thanks, @glnster!

PR #16, opened 2021-07-19
implement handling no func passed into module

glnster:
Address "no function" case.
If no 2nd function arg is passed into module(), assume it is using old syntax of Qunit, which will be handled in a separately with another plugin. Skip it entirely for now.

Testing done:

- this stand-alone repo (added test)
- a simple ember app (added similar test)
- a unit test file in a large, private ember app, that doesn't pass in any function as 2nd arg of module()


__tests__/__fixtures__/one-module-no-module-function-test.js
```diff
@@ -0,0 +1,9 @@
+import { module, test } from 'qunit';
```
scalvert, line 1: I think you're conflating the old syntax of tests with the new. New tests structure should never have a top level module without a second function argument. If it does, that's using the old syntax of QUnit.

scalvert (dismissed):
Let's not mix the old and new syntaxes together, but rather build a second babel plugin to handle the old syntax, as previously discussed.

ckundo (approved):

PR #22, opened 2021-07-28
pass specific test context into getTestMetadata() instead of 'this'

glnster:
Update to use `QUnit.config.current.testEnvironment` instead of `this` to get test's context.
E.g. `let testMetadata = _getTestMetadata(QUnit.config.current.testEnvironment);`

Addresses case where "this" context may change due to arrow function passed into module().

Tested
- in plugin
- in simple ember app with arrow funcs, no hooks
- in private ember app

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -158,13 +158,10 @@ function getTestMetadataAssignment(state, t) {
  * @returns Babel variable declaration
  */
 function getTestMetadataDeclaration(state, t) {
-  const getContextCallExpression = t.callExpression(
-    t.identifier(state.opts.getContextUID.name),
-    []
-  );
+  const getContextIdentifier = state.opts.getContextIdentifier;
   const getTestMetadataExpression = t.callExpression(
     t.identifier(state.opts.getTestMetadataUID.name),
-    [getContextCallExpression]
+    [getContextIdentifier]
   );
 
   return t.variableDeclaration('let', [
@@ -212,7 +209,7 @@ function addMetadata({ types: t }) {
     visitor: {
       Program(babelPath, state) {
         const GET_TEST_METADATA = 'getTestMetadata';
-        const GET_CONTEXT = 'getContext';
+        const GET_CONTEXT = 'QUnit.config.current.testEnvironment';
         const { filename } = state.file.opts;
         state.opts.shouldLoadFile = shouldLoadFile(filename);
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -208,6 +209,7 @@ function addMetadata({ types: t }) {
     visitor: {
       Program(babelPath, state) {
         const GET_TEST_METADATA = 'getTestMetadata';
+        const CONTEXT = 'QUnit.config.current.testEnvironment';
```
ckundo, line 212: could this ever resolve to undefined? what is the behavior if/when its not defined?

src/index.js
```diff
@@ -158,9 +158,10 @@ function getTestMetadataAssignment(state, t) {
  * @returns Babel variable declaration
  */
 function getTestMetadataDeclaration(state, t) {
+  const contextIdentifier = state.opts.contextIdentifier;
   const getTestMetadataExpression = t.callExpression(
     t.identifier(state.opts.getTestMetadataUID.name),
-    [t.thisExpression()]
+    [contextIdentifier]
```
ckundo, line 164: is it possible to us `t.identifier('QUnit.config.current.testEnvironment')` here, and remove part of the diff below? that way we wouldn't need to store more properties on `state`

src/index.js
```diff
@@ -219,6 +221,9 @@ function addMetadata({ types: t }) {
         state.opts.setupCall;
         state.opts.moduleFunction;
         state.opts.hooksIdentifier;
+        state.opts.getTestMetadataUID =
+          babelPath.scope.generateUidIdentifier(GET_TEST_METADATA);
+        state.opts.contextIdentifier = t.identifier(CONTEXT);
```
ckundo, line 226: is there any reason to store this on state, other than to use it in `getTestMetadataDeclaration`?

ckundo (approved):
good catch! take a look at my comments and let me know. otherwise i don't see any blockers.

src/index.js
glnster: @scalvert would you know off hand if 'QUnit.config.current.testEnvironment' could ever potentially resolve to undefined?

Re: behavior if undefined, I think the plugin can have access to this and can do a check on its value. If it's undefined I think we can set it to 'this', but that sort of negates what this changes is trying to address.

Another option is to set it to the returned value of `getContext()`. In that case we must import `getContext()` from test-helpers. We aren't using getContext() because, in the case where no 'hooks' is passed into module() and this plugin inserts hooks, we get a weakmap error. I don't know why getContext() is undefined in this case.

glnster: Actually, I don't think (or I'm not sure how) we can check `Qunit...` or `getContext()` in this stage, where the plugin is processing transformations. I think we can only check their values during the test's actual runtime.

ckundo:
this is a good incremental step. we should also define our own `beforeEach` but that can be a followon PR. 🚀

PR #23, opened 2021-07-30
Do not skip remaining transpilation

ckundo:
- babelPath.skip() was exiting transpilation early for susequent transformations.
- remove skip and rely on the susequent moduleName conditional to skip transformations for our plugin


glnster (approved):
gtg. tested nested modules in simple app too

PR #25, opened 2021-08-04
move env check to index

glnster:
This change moves the `process.env.BABEL_TEST_METADATA` check directly into `index`, instead of passing it in via plugin options. It circumvents an issue experienced in certain apps within our testing, where passing a plugin option surfaces a "circular structure" error.

Tested in
- this repo
- simple-app
- larger private apps

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -203,14 +203,16 @@ function shouldLoadFile(filename) {
  * @returns Babel plugin object with Program and CallExpression visitors
  */
 function addMetadata({ types: t }) {
+  console.log('env var', process.env.BABEL_TEST_METADATA);
+
+  if (!process.env.BABEL_TEST_METADATA) {
+    return {};
+  }
+
   return {
     name: 'addMetadata',
     visitor: {
       Program(babelPath, state) {
-        if (!state.opts.enabled) {
-          return;
-        }
-
         const GET_TEST_METADATA = 'getTestMetadata';
         const { filename } = state.file.opts;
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -203,14 +203,14 @@ function shouldLoadFile(filename) {
  * @returns Babel plugin object with Program and CallExpression visitors
  */
 function addMetadata({ types: t }) {
+  if (!process.env.BABEL_TEST_METADATA) {
+    return {};
```
ckundo, line 207: why return an object here?

src/index.js
```diff
@@ -203,14 +203,14 @@ function shouldLoadFile(filename) {
...
+  }
+
   return {
     name: 'addMetadata',
     visitor: {
       Program(babelPath, state) {
-        if (!state.opts.enabled) {
```
ckundo, line 210: can we leave this in case we're able to get the options working?

ckundo (approved):

src/index.js
glnster: I think we should leave this out to avoid any confusion about its presence. We can add it back in as a follow up along with a working options implementation.

src/index.js
ckundo: what about returning false?

glnster: The expected return is an object. I confirmed this by trying to return just `return;` in simple-app. The error I get is `[BABEL] unknown: Plugin/Preset did not return an object.` So, we should return an empty object in this case.

drewlee (commented):
LGTM

PR #26, opened 2021-08-04
Revisit state options and simplify features

ckundo:
- serializing the nodes onto opts was causing circular references.
- simplify the plugin by removing some edge case handling.
- maintain the features that are important in the mid-term.


src/index.js
```diff
@@ -266,68 +60,71 @@ function addMetadata({ types: t }) {
        * @param {object} state
        */
       CallExpression(babelPath, state) {
+        if (!state.opts.enabled) return;
         if (!state.opts.shouldLoadFile) return;
+        if (!babelPath.get('callee').isIdentifier({ name: 'module' })) return;
+        if (babelPath.parentPath.parent.type !== 'Program') return;
```
drewlee, line 66: Can't all of these just be in a single conditional statement?

src/index.js
```diff
@@ -266,68 +60,71 @@ function addMetadata({ types: t }) {
...
 
-            const hasModuleFunction = t.isFunction(moduleFunction);
-
-            // If no 2nd arg function is passed into module(), assume it's using old test syntax,
-            // which will be supported separately. Skip entirely for now.
-            if (!hasModuleFunction) {
-              return;
-            }
+        if (!moduleName || !moduleFunction || !t.isFunction(moduleFunction))
+          return;
```
drewlee, line 74: Nitpick, but would prefer braces for the return statement.

glnster (approved):
lgtm

src/index.js
```diff
@@ -211,46 +28,23 @@ function addMetadata({ types: t }) {
...
-
-        if (emberTestHelpersIndex !== -1) {
-          // Append to existing test-helpers import
-          importDeclarations[emberTestHelpersIndex]
-            .get('body')
-            .container.specifiers.push(getTestMetaDataImportSpecifier);
-        } else {
-          const getTestMetaDataImportDeclaration = t.importDeclaration(
-            [getTestMetaDataImportSpecifier],
+        babelPath.unshiftContainer(
```
drewlee, line 41: Do we need to do anything in terms of insuring we're not injecting duplicate imports of `@ember/test-helpers` and/or `getTestMetadata`?

ckundo: we have a test for that case... I'm not 100% sure what the behavior is for duplicate imports, but I assume the second one takes precedence, which is what we want. https://github.com/babel-plugin-ember-test-metadata/babel-plugin-ember-test-metadata/pull/26/files#diff-e6827fdf28f3b2f9dec0318dc552eec4d5db91ed9a7958d5d9048a9d453fe51cR97

drewlee: That might have negative implications for the way modules get resolved in the Ember pipeline. This may be ok for Babel itself, but not so much once all the modules get bundled together and get transpiled into non-ESM module loaders. My point is, the test is not a good indicative of how things get coalesced in an Ember app.

ckundo: sounds good. pushed an update...

src/index.js
```diff
@@ -266,68 +69,75 @@ function addMetadata({ types: t }) {
...
        */
       CallExpression(babelPath, state) {
-        if (!state.opts.shouldLoadFile) return;
+        if (
+          !state.opts.enabled ||
+          !state.opts.shouldLoadFile ||
+          !babelPath.get('callee').isIdentifier({ name: 'module' }) ||
+          babelPath.parentPath.parent.type !== 'Program') {
+          return
+        };
```
drewlee, line 78:
```suggestion
          return;
        }
```

drewlee (changes requested):

src/index.js
```diff
@@ -211,46 +28,32 @@ function addMetadata({ types: t }) {
...
-        state.opts.transformedModules = [];
-        state.opts.setupCall;
-        state.opts.moduleFunction;
-        state.opts.hooksIdentifier;
-        state.opts.getTestMetadataUID =
-          babelPath.scope.generateUidIdentifier(GET_TEST_METADATA);
+        const metadata = babelPath.get('body')
+          .filter((n) => n.type === 'ImportDeclaration')
+          .find(n => {
+            return n.get('specifiers').some(s => s.get('container').container.imported.name === "getTestMetadata");
```
glnster, line 41: Can we use `getNodeProperty` here? Ran into an error getting 'name' from undefined.
```
            return n
              .get('specifiers')
              .some(
                (s) =>
                  getNodeProperty(
                    s.get('container'),
                    'container.imported.name'
                  ) === 'getTestMetadata'
              );
```

drewlee (approved):

PR #28, opened 2021-08-06
Cleanup of file casing, incorrect configs

scalvert:
Cleans up a few untidy pieces:

- removes unnecessary file extension and cache specification in eslint script
- changes target to reflect supported target of node in .babelrc (node 6 is _very_ deprecated)
- changes sourceType to module for eslint (unsure why it was script...)
- renames common files to correct casing


drewlee (approved):
LGTM. Thanks for addressing these!

PR #29, opened 2021-08-31
Support embroider file path references

glnster:
There’s an issue with using this plugin in Embroider-built apps, in that the filePath value that is ultimately assigned to test metadata references a temp directory. For example: ../../../../../../tmp/embroider/07f011/tests/acceptance/application-test.js

This may cause issues for downstream consumers that function on a file path structure based on the project's actual/original (non temp) locations.

## The changes here:

- adds util that checks if `filename` (file path) is an Embroider prefixed path
- adds util that strips the Embroider prefix from the file path. For the example above, the result would be "tests/acceptance/application-test.js"
- adds unit & application tests

## Testing

- yarn test
- Also tested in large Ember + Embroider app

### DIFF
--- a/__tests__/utils-test.js
+++ b/__tests__/utils-test.js
@@ -1,4 +1,8 @@
-const { getNodeProperty, stripEmbroiderPrefix } = require('../src/utils');
+const {
+  getNodeProperty,
+  stripEmbroiderPrefix,
+  hasEmbroiderPrefix,
+} = require('../src/utils');
 
 describe('Unit | utils | getNodeProperty', () => {
   it('returns property as expected', () => {
@@ -33,5 +37,18 @@ describe('Unit | utils | stripEmbroiderPrefix', () => {
     expect(stripEmbroiderPrefix(mockEmbroiderBuildPath)).toBe(
       'tests/acceptance/my-test.js'
     );
+  });
+});
+
+describe('Unit | utils | hasEmbroiderPrefix', () => {
+  const mockEmbroiderBuildPath =
+    '/private/var/folders/abcdefg1234/T/embroider/098765/tests/acceptance/my-test.js';
+
+  it('returns true if file path includes embroider', () => {
+    expect(hasEmbroiderPrefix(mockEmbroiderBuildPath)).toBe(true);
+  });
+
+  it('returns false if file path does not include embroider', () => {
+    expect(hasEmbroiderPrefix('this/is/not-an-embroider/path')).toBe(false);
   });
 });
--- a/__tests__/utils-test.js
+++ b/__tests__/utils-test.js
@@ -1,4 +1,4 @@
-const os = require('os');
+const path = require('path');
 const { getNodeProperty, getNormalizedFilePath } = require('../src/utils');
 
 describe('Unit | utils | getNodeProperty', () => {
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -112,7 +116,12 @@ function addMetadata({ types: t }) {
...
         ]);
 
-        const { root, filename } = state.file.opts;
+        let { root, filename } = state.file.opts;
+
+        if (hasEmbroiderPrefix(filename)) {
+          filename = getEmbroiderStrippedPrefixPath(filename);
+        }
+
         const relativeFilePath = path.relative(root, filename);
```
scalvert, line 125: Rather than do this, I would instead encapsulate the entire logic for getting a filename in a separate util, and test that with all the correct permutations. This does a nicer job encapsulating the logic in a single location, whereas this spreads it out within the call site.

src/utils.js
```diff
@@ -30,4 +30,35 @@ function getNodeProperty(node, path) {
...
+
+/**
+ * Get a file path with the Embroider prefix segments stripped out
+ * @param {string} filepath  E.g. /private/var/folders/abcdefg1234/T/embroider/098765/tests/acceptance/my-test.js
+ * @returns {string} E.g. tests/acceptance/my-test.js
+ */
+function getEmbroiderStrippedPrefixPath(filepath) {
+  if (typeof filepath !== 'string') return;
+
+  const separator = filepath.includes('\\') ? '\\' : '/';
```
scalvert, line 53: Why do you need this? If you're trying to get a platform specific path, you can use [path.sep](https://nodejs.org/api/path.html#path_path_sep).

src/utils.js
```diff
@@ -30,4 +30,35 @@ function getNodeProperty(node, path) {
...
-module.exports = getNodeProperty;
+/**
+ * Check if the file path follows an Embroider build pattern
+ * @param {string} filepath
+ * @returns boolean
+ */
+function hasEmbroiderPrefix(filepath) {
+  if (typeof filepath !== 'string') return;
+
+  const separator = filepath.includes('\\') ? '\\' : '/';
```
scalvert, line 41: Why do you need this? If you're trying to get a platform specific path, you can use [path.sep](https://nodejs.org/api/path.html#path_path_sep).

src/utils.js
```diff
@@ -30,4 +30,35 @@ function getNodeProperty(node, path) {
...
+  return filepath.split(separator).includes('embroider');
```
scalvert, line 42: Why do you have to split it to determine if the path has `embroider` in it?

src/utils.js
```diff
@@ -30,4 +30,35 @@ function getNodeProperty(node, path) {
...
+ * @param {string} filepath  E.g. /private/var/folders/abcdefg1234/T/embroider/098765/tests/acceptance/my-test.js
+ * @returns {string} E.g. tests/acceptance/my-test.js
+ */
+function getEmbroiderStrippedPrefixPath(filepath) {
+  if (typeof filepath !== 'string') return;
+
+  const separator = filepath.includes('\\') ? '\\' : '/';
+  const tokens = filepath.split(separator);
+
+  tokens.splice(0, tokens.lastIndexOf('embroider') + 2);
```
scalvert, line 56: Let's extract the `2` to a variable with a semantic name, something like `RELATIVE_PATH_ROOT`. Having magic numbers in the code makes it really hard for readers to understand what you're intent is, and they're left with inferring it.

__tests__/utils-test.js
```diff
@@ -24,3 +28,45 @@ describe('Unit | utils | getNodeProperty', () => {
...
   });
 });
+
+describe('Unit | utils | getEmbroiderStrippedPrefixPath', () => {
+  const mockEmbroiderBuildPath =
+    '/private/var/folders/abcdefg1234/T/embroider/098765/tests/acceptance/my-test.js';
+  const mockEmbroiderBuildPathTwoEmbroiderTokens =
+    '/private/var/folders/embroider/abcdefg1234/T/embroider/098765/tests/acceptance/my-test.js';
+  const mockWindowsEmbroiderBuildPath =
+    'C:\\private\\var\\folders\\abcdefg1234\\T\\embroider\\098765\\tests\\acceptance\\my-test.js';
```
scalvert, line 38: You don't need to refer to these as mocks. They're just paths that you're passing in.

scalvert (changes requested):

src/utils.js
glnster: This is so I can split on `path.sep` and find "embroider" (whole word) as a file path segment.

__tests__/utils-test.js
```diff
@@ -24,3 +25,51 @@ describe('Unit | utils | getNodeProperty', () => {
...
+    expect(
+      getNormalizedFilePath({
+        root: '',
+        filename: 'this/is/not-an-embroider/path',
+      })
+    ).toBe('this/is/not-an-embroider/path');
+  });
+});
+
+describe('Unit | utils | getNormalizedFilePath with Windows path', () => {
```
scalvert, line 57: This feels odd to do. Why are we doing this vs. just configuring `windows-latest` [in our ci workflow](https://github.com/babel-plugin-ember-test-metadata/babel-plugin-ember-test-metadata/blob/master/.github/workflows/ci-build.yml#L19)?

__tests__/utils-test.js
```diff
@@ -24,3 +25,51 @@ describe('Unit | utils | getNodeProperty', () => {
...
+  const originalPathSeperator = path.sep;
+  const windowsEmbroiderBuildPath =
+    'C:\\private\\var\\folders\\abcdefg1234\\T\\embroider\\098765\\tests\\acceptance\\my-test.js';
+
+  beforeEach(() => {
+    path.sep = '\\';
```
scalvert, line 63: I would not do this. See my comment above.

scalvert (changes requested):

src/utils.js
```diff
@@ -30,4 +32,26 @@ function getNodeProperty(node, path) {
...
+/**
+ * Get a normalized file path. If Embroider prefix is present, strip it out
+ * @param {object} fileOpts Babel state.file.opts which include root and filename props
+ * @returns {string} E.g. tests/acceptance/my-test.js
+ */
+function getNormalizedFilePath(fileOpts) {
+  let { root, filename } = fileOpts;
+  const tokens = filename.split(path.sep);
+
+  if (tokens.includes('embroider')) {
```
ckundo, line 44: `embroider` could be a constant.

__tests__/utils-test.js
ckundo: I think if you built these strings with an array and `path.join`, you'd only need one test since `join` should use the appropriate separator for the current CI platform. do add `windows-latest` to the ci workflow tho so we exercise that code path...

ckundo (commented):
maybe use `path.join` rather than stub the `path` object, and add `windows-latest` to ci workflow.

__tests__/utils-test.js
glnster: I think I may need to setup babel-plugin-tester snap shot tests to somehow dynamically point to one snapshot vs another, based on platform (Win vs *Nix).

glnster: I updated the tests to use path.join.

Adding `windows-latest` to the ci workflow should be follow on work imo, as I may need to create separate output fixtures for Win for each fixture. https://github.com/babel-utils/babel-plugin-tester#outputfixture

ckundo (approved):

scalvert (approved):

scalvert:
One thing we didn't do in this PR is update the CI tests with `windows-latest`. Can we do that, or are you planning on doing that in a separate PR with the updates to point to windows snapshots too?

glnster:
> One thing we didn't do in this PR is update the CI tests with `windows-latest`. Can we do that, or are you planning on doing that in a separate PR with the updates to point to windows snapshots too?

I plan to do this as a separate PR. I created an issue #30 to track this.

PR #31, opened 2021-09-01
Adds window-latest to test matrix for CI

scalvert:


scalvert:
Using `outputFixture` works in debug mode, but for some reason Jest tries to require that file in non-debug mode, which causes errors due to missing dependencies.

scalvert:
Going to close this for now. We can add it back if/when we need it.

PR #33, opened 2021-09-08
Converting to monorepo

scalvert:
Converts this repository to use a yarn workspaces monorepo. This prepares us for adding acceptance-style tests for variants of ember apps.


packages/babel-plugin-ember-test-metadata/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "lint-staged": {
+    "**/src/*.js": [
+      "prettier --write"
+    ]
+  },
+  "prettier": {
+    "bracketSpacing": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
```
rwjblue, line 38: I think you'll want to copy this to be in the top level `package.json` too (this is why the top level `.eslintrc.js` needed single quote -> double quote changes)

.eslintrc.js
```diff
@@ -3,15 +3,15 @@ module.exports = {
     browser: false,
     node: true,
   },
-  plugins: ['node', 'prettier'],
-  extends: ['eslint:recommended', 'plugin:node/recommended', 'prettier'],
+  plugins: ["node", "prettier"],
+  extends: ["eslint:recommended", "plugin:node/recommended", "prettier"],
```
rwjblue, line 7: Seems odd to have this differ from the rest of the project, I think it's doing this because of missing prettier config in the root

packages/babel-plugin-ember-test-metadata/package.json
scalvert: yep good call

.eslintrc.js
scalvert: confirm

packages/babel-plugin-ember-test-metadata/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "author": {
+    "name": "Glenn Cueto",
+    "email": "gcueto@linkedin.com"
+  },
+  "main": "dist/index.js",
+  "files": [
+    "dist/*.js"
+  ],
+  "scripts": {
+    "prepublishOnly": "npm run build",
```
ckundo, line 17: are we using `npm` in this project?

packages/babel-plugin-ember-test-metadata/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "prebuild": "rm -rf dist",
+    "build": "babel src --out-dir dist --copy-files --ignore **/*.test.js",
+    "postbuild": "rm -rf dist/**/*.test.js",
```
ckundo, line 20: this feels odd... do we need to transpile tests into dist to begin with?

ckundo (approved):
curious about some of our package.json scripts, but otherwise good

packages/babel-plugin-ember-test-metadata/package.json
scalvert: Nope. Though it doesn't really matter in this case, since it's just running a build. I'll change it to yarn (though it was like this previously).

packages/babel-plugin-ember-test-metadata/package.json
scalvert: This isn't any different than what we did before. If we want to change this infra, let's keep this separate from the monorepo changes. Happy to look at this again after.

PR #37, opened 2021-09-11
DRAFT Fix embroider tests

glnster:
Sets up the embroider-app scenario's prepared path and updates embroider tests to match on the exact expected path for all tests.

Before: Embroider's build step was building into a generated path like this: `/private/var/folders/2j/wk59f3qx2kl7zl34yr8yhn38000j2y/T/embroider/7eee16/private/var/folders/2j/wk59f3qx2kl7zl34yr8yhn38000j2y/T/tmp-72329-6SsJEomJQDVe`

After: the embroider scenario is still prepared in a tmp directory and embroider file path segments are appended like: `/private/var/folders/2j/wk59f3qx2kl7zl34yr8yhn38000j2y/T/embroider/7eee16/private/var/folders/2j/wk59f3qx2kl7zl34yr8yhn38000j2y/T/tmp-72329-6SsJEomJQDVe/embroider/999999`

TODO (in a separate PR)

- Branch, and update the CLASSIC tests to match on the exact expected path for all tests (eg. tests/unit/some-test.js)

### DIFF
--- a/packages/test-scenarios/__tests__/scenario-test.js
+++ b/packages/test-scenarios/__tests__/scenario-test.js
@@ -14,6 +14,59 @@ function createTmpDir() {
 
 async function classic(project) {
   merge(project.files, {
+    tests: {
+      unit: {
+        'with-hooks-test.js': `import { module, test } from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
+
+module('Acceptance | with-hooks-test', function (hooks) {
+hooks.beforeEach(function () {
+  // noop
+});
+
+test('example', async function (assert) {
+  assert.ok(getTestMetadata(this).filePath.includes('tests/unit/with-hooks-test.js'));
+});
+});
+`,
+        'without-hooks-test.js': `import { module, test } from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
+
+module('Acceptance | without-hooks-test', function () {
+hooks.beforeEach(function () {
+  // noop
+});
+
+test('example', async function (assert) {
+  assert.ok(getTestMetadata(this).filePath.includes('tests/unit/without-hooks-test.js'));
+});
+});
+`,
+        'with-multiple-modules-test.js': `import {module, test} from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
+
+module('Acceptance | with-multiple-modules-test', function (hooks) {
+hooks.beforeEach(function () {
+  // noop
+});
+
+test('example', async function (assert) {
+  assert.ok(getTestMetadata(this).filePath.includes('tests/unit/with-multiple-modules-test.js'));
+});
+});
+
+module('Acceptance | with-multiple-modules-test 2', function (hooks) {
+hooks.beforeEach(function () {
+  // noop
+});
+
+test('example', async function (assert) {
+  assert.ok(getTestMetadata(this).filePath.includes('tests/unit/with-multiple-modules-test.js'));
+});
+});
+`,
+      },
+    },
     'ember-cli-build.js': `'use strict';
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 module.exports = function (defaults) {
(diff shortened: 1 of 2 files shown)


packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -1,15 +1,74 @@
 const { Scenarios, Project } = require('scenario-tester');
-const { dirname } = require('path');
+const { dirname, join } = require('path');
 const { merge } = require('lodash');
+const { realpathSync } = require('fs-extra');
+const tmp = require('tmp');
 
 jest.setTimeout(500000);
 
+const rootTmpDir = createTmpDir();
```
ckundo, line 9: should we should add this in a `before` block and remove in an `after`?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -1,15 +1,74 @@
...
+  return realpathSync(tmp.dirSync({ unsafeCleanup: true }).name);
+}
+
 async function classic(project) {
   merge(project.files, {
-    'ember-cli-build.js': `'use strict';
+    tests: {
+      unit: {
+        'with-hooks-test.js': `import { module, test } from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
```
ckundo, line 20: could these template strings be extracted into a fixture? it might make it easier to reason about the intent of this setup.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -85,69 +196,20 @@ Scenarios.fromProject(baseApp)
...
-  });
-});
-`,
-        },
-      },
-    });
   })
   .forEachScenario((scenario) => {
     describe(scenario.name, () => {
+      const EMBROIDER = 'embroider';
```
ckundo, line 202: I'd put these constants at the top of the file, since they're static.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -85,69 +196,20 @@ Scenarios.fromProject(baseApp)
...
+      const EMBROIDER_PATH_SEGMENT = '999999';
       let app;
+      let appTmp;
 
       beforeAll(async () => {
-        app = await scenario.prepare();
+        if (scenario.name.includes('embroider-app')) {
```
ckundo, line 208: this looks brittle. we should be able to change the scenario name without breaking this test.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -85,69 +196,20 @@ Scenarios.fromProject(baseApp)
...
-});
-`,
-        },
-      },
-    });
   })
   .forEachScenario((scenario) => {
     describe(scenario.name, () => {
+      const EMBROIDER = 'embroider';
+      const EMBROIDER_PATH_SEGMENT = '999999';
```
ckundo, line 203: maybe we could generate the SHA here instead of making it constant. that'd be a higher fidelity test case I think.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -85,69 +196,20 @@ Scenarios.fromProject(baseApp)
...
       let app;
+      let appTmp;
 
       beforeAll(async () => {
-        app = await scenario.prepare();
+        if (scenario.name.includes('embroider-app')) {
+          appTmp = join(rootTmpDir, EMBROIDER, EMBROIDER_PATH_SEGMENT);
+        }
+
+        app = await scenario.prepare(appTmp);
```
ckundo, line 212: what happens when `appTmp` is undefined?

packages/test-scenarios/__tests__/scenario-test.js
ckundo: maybe we could have an enum object with fixed keys for the scenario names. e.g.

```
const SCENARIOS = {
  embroider: "embroider-app"
}
```

you would then use this lookup in both the scenario definition and here.

packages/test-scenarios/__tests__/scenario-test.js
glnster: It falls back to using the project.baseDir tmp directory https://github.com/ef4/scenario-tester/blob/main/index.ts#L151

packages/test-scenarios/__tests__/scenario-test.js
scalvert: Already doing that now in a separate branch.

scalvert:
Can we pause on this PR? This PR is duplicating work that is superseded by some changes I am currently working on.

glnster:
> Can we pause on this PR? This PR is duplicating work that is superseded by some changes I am currently working on.

Yes. I paused on this and am working on fixing for classic build file paths.

PR #38, opened 2021-09-13
Add in repo test scenarios

scalvert:
Adds tests for in-repo addon scenarios. Specifically, this handles the scenarios where we have in-repo addons with colocated tests, and the app uses `ember-add-in-repo-tests`. 

We only need to cover the in-repo addon scenario and not in-repo engines as the latter is virtually identical to the addon case, and is redundant.

### DIFF
--- a/packages/test-scenarios/__tests__/scenario-test.js
+++ b/packages/test-scenarios/__tests__/scenario-test.js
@@ -1,5 +1,6 @@
 const { Scenarios, Project } = require('scenario-tester');
-const { dirname } = require('path');
+const { dirname, join } = require('path');
+const { readFileSync } = require('fs');
 const { merge } = require('lodash');
 
 jest.setTimeout(500000);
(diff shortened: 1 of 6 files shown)


rwjblue (approved):

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -66,6 +72,47 @@ module.exports = function (defaults) {
...
+  await addInRepoAddon(project, 'fake-addon');
+}
+
+async function addInRepoAddon(project, name, version = '0.0.0') {
+  project.linkDependency('ember-add-in-repo-tests', {
+    baseDir: __dirname,
+  });
+
+  project.pkg['ember-addon'] = {
+    paths: [`lib/${name}`],
```
glnster, line 86: will we need to use join to generate these paths?

glnster (approved):

packages/test-scenarios/__tests__/scenario-test.js
scalvert: We could, but our apps/tests are all assuming *nix right now. I know we discussed windows support, but we should consider adding that when needed.

PR #39, opened 2021-09-13
Fix file paths for CLASSIC Ember builds

glnster:
Updates the plugin to provide correct, normalized file paths for CLASSIC Ember builds.
Separates out the test fixtures for the other scenarios, for now.
Adds unit tests for the new functions used for parsing `defaults.project` and getting parsed file paths.

Addressing the classic in-repo addon build scenario will be in a follow on PR.

### DIFF
--- a/packages/babel-plugin-ember-test-metadata/__tests__/index-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/index-test.js
@@ -2,7 +2,7 @@ const pluginTester = require('babel-plugin-tester').default;
 const addMetadata = require('../index');
 const path = require('path');
 
-const projectInfo = {
+const projectConfiguration = {
   pkg: {
     name: 'packages/babel-plugin-ember-test-metadata',
     'ember-addon': {},
@@ -13,7 +13,7 @@ pluginTester({
   plugin: addMetadata,
   pluginName: 'addMetadata',
   snapshot: true,
-  pluginOptions: { enabled: true, projectInfo },
+  pluginOptions: { enabled: true, projectConfiguration },
   tests: [
     {
       title: 'for a module with hooks it adds getTestMetadata',
--- a/packages/babel-plugin-ember-test-metadata/__tests__/index-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/index-test.js
@@ -2,11 +2,18 @@ const pluginTester = require('babel-plugin-tester').default;
 const addMetadata = require('../index');
 const path = require('path');
 
+const projectInfo = {
+  pkg: {
+    name: 'packages/babel-plugin-ember-test-metadata',
+    'ember-addon': {},
+  },
+};
+
 pluginTester({
   plugin: addMetadata,
   pluginName: 'addMetadata',
   snapshot: true,
-  pluginOptions: { enabled: true },
+  pluginOptions: { enabled: true, projectInfo },
   tests: [
     {
       title: 'for a module with hooks it adds getTestMetadata',
--- a/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
@@ -96,34 +96,26 @@ describe('getNormalizedFilePath', () => {
 });
 
 describe('_getParsedClassicFilepath', () => {
-  const projectConfiguration = {
-    pkg: {
-      name: 'test-app',
-      'ember-addon': {},
-    },
-  };
-  const pathSegments = ['test-app', 'tests', 'acceptance', 'my-test.js'];
-  const monoRepoProjectConfiguration = {
-    pkg: {
-      name: `@parent-repo-name${path.sep}test-app`,
-      'ember-addon': {},
-    },
-  };
-  const monoRepoPathSegments = [
-    '@parent-repo-name',
-    'test-app',
-    'tests',
-    'acceptance',
-    'my-test.js',
-  ];
-
   it('returns file path from classic build correctly', () => {
-    expect(_getParsedClassicFilepath(pathSegments, projectConfiguration)).toBe(
-      path.join('tests', 'acceptance', 'my-test.js')
-    );
-    expect(_getParsedClassicFilepath(monoRepoPathSegments, monoRepoProjectConfiguration)).toBe(
-      path.join('tests', 'acceptance', 'my-test.js')
-    );
+    expect(
+      _getParsedClassicFilepath(['test-app', 'tests', 'acceptance', 'my-test.js'], {
+        pkg: {
+          name: 'test-app',
+          'ember-addon': {},
+        },
+      })
+    ).toBe(path.join('tests', 'acceptance', 'my-test.js'));
+    expect(
+      _getParsedClassicFilepath(
+        ['@parent-repo-name', 'test-app', 'tests', 'acceptance', 'my-test.js'],
+        {
+          pkg: {
+            name: `@parent-repo-name${path.sep}test-app`,
+            'ember-addon': {},
+          },
+        }
+      )
+    ).toBe(path.join('tests', 'acceptance', 'my-test.js'));
   });
 });
 
--- a/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
@@ -64,18 +64,24 @@ describe('Unit | utils | getNormalizedFilePath', () => {
       filename: path.join('this', 'is', 'not-an-embroider', 'path'),
     },
   };
+  const projectInfo = {
+    pkg: {
+      name: 'test-app',
+      'ember-addon': {},
+    },
+  };
 
   it('returns stripped file path as expected', () => {
-    expect(getNormalizedFilePath(fileOpts.embroiderBuildPath)).toBe(
+    expect(getNormalizedFilePath(fileOpts.embroiderBuildPath, projectInfo)).toBe(
       path.join('tests', 'acceptance', 'my-test.js')
     );
-    expect(getNormalizedFilePath(fileOpts.embroiderBuildPathTwoEmbroiderTokens)).toBe(
+    expect(getNormalizedFilePath(fileOpts.embroiderBuildPathTwoEmbroiderTokens, projectInfo)).toBe(
       path.join('tests', 'acceptance', 'my-test.js')
     );
   });
 
   it('returns unmodified file path when path does not include "embroider" as a segment', () => {
-    expect(getNormalizedFilePath(fileOpts.normalFilePath)).toBe(
+    expect(getNormalizedFilePath(fileOpts.normalFilePath, projectInfo)).toBe(
       path.join('this', 'is', 'not-an-embroider', 'path')
     );
   });
--- a/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
@@ -1,5 +1,10 @@
 const path = require('path');
-const { getNodeProperty, getNormalizedFilePath } = require('../utils');
+const {
+  getNodeProperty,
+  getNormalizedFilePath,
+  getParsedClassicFilepath,
+  getParsedEmbroiderFilepath,
+} = require('../utils');
 
 describe('Unit | utils | getNodeProperty', () => {
   it('returns property as expected', () => {
@@ -84,5 +89,51 @@ describe('Unit | utils | getNormalizedFilePath', () => {
     expect(getNormalizedFilePath(fileOpts.normalFilePath, projectInfo)).toBe(
       path.join('this', 'is', 'not-an-embroider', 'path')
     );
+  });
+});
+
+describe('Unit | utils | getParsedClassicFilepath', () => {
+  const projectInfo = {
+    pkg: {
+      name: 'test-app',
+      'ember-addon': {},
+    },
+  };
+  const tokens = ['test-app', 'tests', 'acceptance', 'my-test.js'];
+  const monoRepoProjectInfo = {
+    pkg: {
+      name: `@parent-repo-name${path.sep}test-app`,
+      'ember-addon': {},
+    },
+  };
+  const monoRepoTokens = ['@parent-repo-name', 'test-app', 'tests', 'acceptance', 'my-test.js'];
+
+  it('returns file path from classic build correctly', () => {
+    expect(getParsedClassicFilepath(tokens, projectInfo)).toBe(
+      path.join('tests', 'acceptance', 'my-test.js')
+    );
+    expect(getParsedClassicFilepath(monoRepoTokens, monoRepoProjectInfo)).toBe(
+      path.join('tests', 'acceptance', 'my-test.js')
+    );
+  });
+});
+
+describe('Unit | utils | getParsedEmbroiderFilepath', () => {
+  const filename = path.join(
+    'private',
+    'var',
+    'folders',
+    'abcdefg1234',
+    'T',
+    'embroider',
+    '098765',
+    'tests',
+    'acceptance',
+    'my-test.js'
+  );
+  const tokens = filename.split(path.sep);
+
+  it('returns file path from classic build correctly', () => {
+    expect(getParsedEmbroiderFilepath(tokens)).toBe(path.join('tests', 'acceptance', 'my-test.js'));
   });
 });
--- a/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
+++ b/packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
@@ -2,11 +2,11 @@ const path = require('path');
 const {
   getNodeProperty,
   getNormalizedFilePath,
-  getParsedClassicFilepath,
-  getParsedEmbroiderFilepath,
+  _getParsedClassicFilepath,
+  _getParsedEmbroiderFilepath,
 } = require('../utils');
 
-describe('Unit | utils | getNodeProperty', () => {
+describe('getNodeProperty', () => {
   it('returns property as expected', () => {
     const mockNode = {
       system: {
@@ -31,7 +31,7 @@ describe('Unit | utils | getNodeProperty', () => {
   });
 });
 
-describe('Unit | utils | getNormalizedFilePath', () => {
+describe('getNormalizedFilePath', () => {
   const fileOpts = {
     embroiderBuildPath: {
       root: '',
@@ -48,7 +48,7 @@ describe('Unit | utils | getNormalizedFilePath', () => {
         'my-test.js'
       ),
     },
-    embroiderBuildPathTwoEmbroiderTokens: {
+    embroiderBuildPathTwoEmbroiderPathSegments: {
       root: '',
       filename: path.join(
         'private',
@@ -69,7 +69,7 @@ describe('Unit | utils | getNormalizedFilePath', () => {
       filename: path.join('this', 'is', 'not-an-embroider', 'path'),
     },
   };
-  const projectInfo = {
+  const projectConfiguration = {
     pkg: {
       name: 'test-app',
       'ember-addon': {},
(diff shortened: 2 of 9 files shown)


packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -32,21 +32,44 @@ function getNodeProperty(node, path) {
...
 
   if (tokens.includes(EMBROIDER)) {
     const RELATIVE_PATH_ROOT = 2;
 
     tokens.splice(0, tokens.lastIndexOf(EMBROIDER) + RELATIVE_PATH_ROOT);
   }
+
   filename = tokens.join(path.sep);
 
   return path.relative(root, filename);
```
scalvert, line 75: I'd probably try to split this functionality up so it's not trying to do both in the same function. I'd have a parent function that's called and splits off to do either the classic thing or the embroider thing. That way you can test each independently of one another.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -29,81 +33,81 @@ module.exports = function (defaults) {
...
-          name: require("./package").name,
-        };`,
-        tests: {
-          unit: getTestFiles('with-hooks-test.js'),
-        },
-      },
-    },
-  });
-}
+// async function classicInRepoAddon(project) {
```
scalvert, line 36: Just a preference thing, but I think it's fine to keep these uncommented and just let them run red 🤷‍♂️

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
   const tokens = filename.split(path.sep);
-  const EMBROIDER = 'embroider';
+  const isEmbroider = tokens.includes(EMBROIDER);
 
-  if (tokens.includes(EMBROIDER)) {
-    const RELATIVE_PATH_ROOT = 2;
-
-    tokens.splice(0, tokens.lastIndexOf(EMBROIDER) + RELATIVE_PATH_ROOT);
+  if (isEmbroider) {
+    filename = getParsedEmbroiderFilepath(tokens);
```
ckundo, line 95: not sure if it matters, but this might be easier to reason about this if it were inlined.

```javascript
const { filename } = fileOpts;
// const embroiderBuild = ???
// const classicBuild = ???

let normalizedPath = filename;

if (embroiderBuild) {
  const pathParts = filename.split(path.sep);
  const extendedParts = 2;
  const separatorStart = pathParts.lastIndexOf('embroider');
  pathParts.splice(0, separatorStart + extendedParts)
  normalizedPath = path.join(pathParts);
} else if (classicBuild) {
  const { name } = projectInfo.pkg;
  normalizedPath = normalizedPath.slice(name.length);
}

return path.relative(root, filename);
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
-  filename = tokens.join(path.sep);
 
   return path.relative(root, filename);
 }
 
 module.exports = {
   getNodeProperty,
   getNormalizedFilePath,
+  getProjectInfo,
+  getParsedClassicFilepath,
```
ckundo, line 107: I don't think these actually need to be exported, since they're only used in this file, right? they're effectively private methods.

ckundo (commented):
I'm a little fuzzy on what "classic" means vs "embroider", but thats maybe my lack of understanding. will wait on @scalvert to give the final thumbsup.

packages/babel-plugin-ember-test-metadata/__tests__/index-test.js
```diff
@@ -2,11 +2,18 @@ const pluginTester = require('babel-plugin-tester').default;
 const addMetadata = require('../index');
 const path = require('path');
 
+const projectInfo = {
```
scalvert, line 5: I'd probably call this `pluginConfiguration`, as that's a little more accurate to what it is.

```suggestion
const projectConfiguration = {
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
       filename: path.join('this', 'is', 'not-an-embroider', 'path'),
     },
   };
+  const projectInfo = {
```
scalvert, line 72:
```suggestion
  const projectConfiguration = {
```

scalvert: If you accept this suggestion, make sure you update all the references below!

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
 
   it('returns unmodified file path when path does not include "embroider" as a segment', () => {
-    expect(getNormalizedFilePath(fileOpts.normalFilePath)).toBe(
+    expect(getNormalizedFilePath(fileOpts.normalFilePath, projectInfo)).toBe(
       path.join('this', 'is', 'not-an-embroider', 'path')
     );
   });
 });
+
+describe('Unit | utils | getParsedClassicFilepath', () => {
```
scalvert, line 95: This is a huge nit, but can we not use these types of strings for our suite names? This is very much an Ember/custom convention, not one we should use elsewhere. I think just doing

```suggestion
describe('getParsedClassicFilepath', () => {
```

would be fine.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+  const projectInfo = {
```
scalvert, line 96:
```suggestion
  const projectConfiguration = {
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+    pkg: {
+      name: 'test-app',
+      'ember-addon': {},
+    },
+  };
+  const tokens = ['test-app', 'tests', 'acceptance', 'my-test.js'];
```
scalvert, line 102:
```suggestion
  const pathSegments = ['test-app', 'tests', 'acceptance', 'my-test.js'];
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+  const monoRepoProjectInfo = {
+    pkg: {
+      name: `@parent-repo-name${path.sep}test-app`,
+      'ember-addon': {},
+    },
+  };
+  const monoRepoTokens = ['@parent-repo-name', 'test-app', 'tests', 'acceptance', 'my-test.js'];
```
scalvert, line 109:
```suggestion
  const monoRepoPathSegments = ['@parent-repo-name', 'test-app', 'tests', 'acceptance', 'my-test.js'];
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+
+  it('returns file path from classic build correctly', () => {
+    expect(getParsedClassicFilepath(tokens, projectInfo)).toBe(
```
scalvert, line 112:
```suggestion
    expect(getParsedClassicFilepath(pathSegments, projectInfo)).toBe(
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+      path.join('tests', 'acceptance', 'my-test.js')
+    );
+    expect(getParsedClassicFilepath(monoRepoTokens, monoRepoProjectInfo)).toBe(
```
scalvert, line 115:
```suggestion
    expect(getParsedClassicFilepath(monoRepoPathSegments, monoRepoProjectInfo)).toBe(
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,71 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+    'folders',
+    'abcdefg1234',
+    'T',
+    'embroider',
+    '098765',
+    'tests',
+    'acceptance',
+    'my-test.js'
+  );
+  const tokens = filename.split(path.sep);
```
scalvert, line 134:
```suggestion
  const pathSegments = filename.split(path.sep);
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path');
+const EMBROIDER = 'embroider';
```
scalvert, line 2: Why do we need this to be a const?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
+    parsedProjectInfo.pkg = {
+      name: project.pkg.name,
+      'ember-addon': {
+        paths: project.pkg['ember-addon'].paths,
+      },
+    };
+  }
+
+  return parsedProjectInfo;
+}
```
scalvert, line 54:
```suggestion
function getProjectConfiguration(project) {
  const parsedProjectConfiguration = {};

  if (project) {
    parsedProjectConfiguration.pkg = {
      name: project.pkg.name,
      'ember-addon': {
        paths: project.pkg['ember-addon'].paths,
      },
    };
  }

  return parsedProjectInfo;
}
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
 }
 
 /**
- * Get a normalized file path. If Embroider prefix is present, strip it out
+ * Function to parse defaults.project and only return info to be used by the plugin
+ * @param {object} project Ember defaults.project
+ * @returns {object} Contains project name and ember-addon path info
+ */
```
scalvert, line 40: You don't really need a jsdoc for this, as it's not public. Jsdocs are really for public things that you want consumers to be able to understand. In this case, we just export it to test it, but the only thing public is the babel plugin function.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
+ * @param {object} projectInfo Contains project name, ember-addon path info
+ * @returns {string} A joined file path, e.g. tests/acceptance/my-test.js
+ */
+function getParsedClassicFilepath(tokens, projectInfo) {
+  const projectNameTokens = projectInfo.pkg.name.split(path.sep);
+
+  tokens.splice(0, tokens.indexOf(projectNameTokens[0]) + projectNameTokens.length);
+
+  return tokens.join(path.sep);
+}
```
scalvert, line 68:
```suggestion
function getParsedClassicFilepath(pathSegments, projectConfiguration) {
  const projectNamePathSeparators = projectConfiguration.pkg.name.split(path.sep);

  pathSegments.splice(0, pathSegments.indexOf(projectNamePathSeparators[0]) + projectNamePathSeparators.length);

  return pathSegments.join(path.sep);
}
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
+ * @param {array} tokens File path segments
+ * @returns {string} A joined file path, e.g. tests/acceptance/my-test.js
+ */
+function getParsedEmbroiderFilepath(tokens) {
+  const RELATIVE_PATH_ROOT = 2;
+
+  tokens.splice(0, tokens.lastIndexOf(EMBROIDER) + RELATIVE_PATH_ROOT);
+
+  return tokens.join(path.sep);
+}
```
scalvert, line 81:
```suggestion
function getParsedEmbroiderFilepath(pathSegments) {
  const RELATIVE_PATH_ROOT = 2;

  pathSegments.splice(0, pathSegments.lastIndexOf(EMBROIDER) + RELATIVE_PATH_ROOT);

  return pathSegments.join(path.sep);
}
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,76 @@ function getNodeProperty(node, path) {
...
  * @param {object} fileOpts Babel state.file.opts which include root and filename props
+ * @param {object} projectInfo Contains project name, ember-addon path info
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
-function getNormalizedFilePath(fileOpts) {
+function getNormalizedFilePath(fileOpts, projectInfo) {
   let { root, filename } = fileOpts;
   const tokens = filename.split(path.sep);
-  const EMBROIDER = 'embroider';
+  const isEmbroider = tokens.includes(EMBROIDER);
```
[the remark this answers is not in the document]
scalvert: How is it buggy? Can you elaborate on this?

That path you showed would never exist, since we have intimate knowledge of that build system and the paths it produces. You'd never have a nested reference to `embroider` inside a directory, and if you do, that's bonkers and you deserve to have your stuff break :)

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: I gave the exact opposite advice. Combining both is harder to reason about IMO. I do think it's fine to have a single entry point, and to split up the logic between those in separate functions. That's what I'd indicated in my prior comment.

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: If they're tested (which they should be, they need to be exported).

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/utils.js
glnster: Yes, they are tested, so yes, needed to export.

glnster: Would it be weird to prefix them with an underscore, since they are private, and export them as such, for testing?

packages/babel-plugin-ember-test-metadata/utils.js
glnster: This string is used in 2 different places.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: This specific function is public. It is called in `ember-cli-build`. https://github.com/babel-plugin-ember-test-metadata/babel-plugin-ember-test-metadata/blob/9172371687e0d7b64edd79d903cfda3f9bfbc94d/packages/test-scenarios/__tests__/scenario-test.js#L23

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: It wouldn't be weird at all, though I'm not sure that's necessary. They're in a utils module which someone would have to dig for if they wanted to use them. Also, does it really matter?

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I appreciate both perspectives. In this case, splitting them out to smaller functions made it more readable to me, easier to grok, and unit-testable. I think eventually we will have a 3rd function that these functions will use, that will be for handling in-repo addon cases.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -64,19 +69,82 @@ describe('Unit | utils | getNormalizedFilePath', () => {
...
+      'ember-addon': {},
+    },
+  };
+  const monoRepoPathSegments = [
+    '@parent-repo-name',
+    'test-app',
+    'tests',
+    'acceptance',
+    'my-test.js',
+  ];
```
scalvert, line 118: Wondering what the benefit of these constants is vs just having the assertion arrangement happening right in the test. If you do it outside, you force the reader of the test to look around and read all the pieces of the test in order to understand what it's doing.

The way to think about tests are, they're usually write once (or few!) and read many. That means people use the tests as a reference to the expected behavior of the code. If you optimize for the write case, that can come as an expense to the read case.

ckundo (approved):

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: What's the problem with having a string in two places? It's more readable to inline this, and I'm not a fan of creating string constants just because they're used in more than one place except if they're of sufficient complexity that they'd need to be done once to avoid bugs. In this case I'm not sure it fits that.

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: Thanks for pointing that out. Let's document it properly then. Any objects that have specific sub-properties need to be documented using the dot notation.

Example:

```js
/**
 * Function to parse defaults.project and only return info to be used by the plugin
 * @param {object} project - Ember defaults.project.
 * @param {string} project.pkg - The package of the parent project.
 * @param {string} project.pkg.name - The name of the parent project.
 * @returns {object} Contains project name and ember-addon path info
 */
```

The above is just an example, but should give you an idea of how to document sub-properties.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +34,69 @@ function getNodeProperty(node, path) {
...
+    pathSegments.indexOf(projectNamePathSeparators[0]) + projectNamePathSeparators.length
+  );
+
+  return pathSegments.join(path.sep);
+}
+
+function _getParsedEmbroiderFilepath(pathSegments) {
+  const RELATIVE_PATH_ROOT = 2;
+
+  // Strip out Embroider prefix (embroider/nnnnnn) segments from file path
```
scalvert, line 70: If we're going to put a comment here, let's make it more complete. Let's indicate what a typical `pathSegments` array would look like to better indicate what we're doing to it.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
glnster: Thanks for the wisdom shared. I've inlined them and it is more readable now, what the specific test is doing.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I went ahead and gave them underscores

packages/babel-plugin-ember-test-metadata/utils.js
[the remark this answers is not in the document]
ckundo: it's brittle because its coupling to something even more transient than a private API, its bound essentially to an internal constant string that the maintainer could change without any expectation that it'd break things. certainly we can say that should never happen, but that doesn't make it less brittle.

packages/babel-plugin-ember-test-metadata/utils.js
ckundo: IMO, tests should exercise the public API; corollary there is that if you are testing it, it is in the public API.

packages/babel-plugin-ember-test-metadata/utils.js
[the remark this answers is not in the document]
scalvert: It's a consistent part of the build system for embroider. I do acknowledge that a env var or something similar could be better, but in lieu of that this seems reasonable. Also, we've run this past those who are building embroider and this was the recommended approach.

If we feel really strongly about this, we can also pass a boolean value in to the plugin to indicate whether it's using embroider or not. It can easily introspect to determine whether the app is using embroider since the module where we'd pass the options into this babel plugin would be the same module where embroider APIs are being used. WDYT?

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: So you're saying no private API should be tested? That means we should only be testing at the higher level, essentially acceptance testing?

FWIW, this pattern is used heavily throughout a lot of our node/Ember ecosystem. I can point you to references if you'd like.

ckundo: unless there's undue difficulty in exercising the internals of a public API, then yes, I avoid testing private methods. There are advantages to keeping private methods private, and as a result this is not a unique or terribly controversial opinion, but I'm not rigid about it either.

packages/babel-plugin-ember-test-metadata/utils.js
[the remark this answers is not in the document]
ckundo: that makes sense and thanks for the info. this is a special case in a way, and I'm happy to take the recommended approach.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: Sync'd with Cam during our regular sync and we are ok to close this convo point. Appreciate all the input and insightful perspective.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -33,26 +33,75 @@ function getNodeProperty(node, path) {
...
 
-  if (tokens.includes(EMBROIDER)) {
-    const RELATIVE_PATH_ROOT = 2;
-
-    tokens.splice(0, tokens.lastIndexOf(EMBROIDER) + RELATIVE_PATH_ROOT);
+  if (isEmbroider) {
+    filename = _getParsedEmbroiderFilepath(pathSegments);
+  } else {
+    filename = _getParsedClassicFilepath(pathSegments, projectConfiguration);
   }
```
scalvert, line 96:
```suggestion
    filename = isEmbroider ? _getParsedEmbroiderFilepath(pathSegments) : _getParsedClassicFilepath(pathSegments, projectConfiguration);
```

scalvert (approved):

PR #40, opened 2021-09-14
Fix file paths & scenarios for Classic in-repo addon, Embroider, and Embroider + addon

glnster:
## Summary
Updates the plugin to provide correct, normalized file paths for:

- Classic in-repo addon Ember builds and `ember-add-in-repo-tests` scenario
- Embroider build scenario
- Embroider + addon scenario

### utils.js
- Refactored the normalization implementation to use `path` API instead of arrays for path manipulation. This led to removing parsing functions no longer needed and updating the existing `getNormalizedFilePath` function.
- Added a new `addonFound` function to support addon scenarios.

### utils-test.js
- Remove tests for removed functions.
- Inline some test config instead of storing them as const.

### test fixtures
- Updated our existing scenario test fixtures with a placeholder string. The string is updated dynamically at test scenario run time so that we can assert for classic/embroider/addon file paths. E.g. for addon scenarios, we assert a file path that starts as "lib/addon-name/tests/acceptance/..." vs. classic/embroider file paths that start as "tests/acceptance/..."

### scenario-test.js
- In each scenario, set `project.pkg.scenarioTesterTests`. This config is used for merging test files into the specified dir structure for that scenario.
- Added a `classicMultipleInRepoAddon` scenario to test the multiple-addon case.
- Updated the Embroider scenario with our `getProjectConfiguration` config.
- Updated the `addInRepoAddon` function with proper addon config.
- Extended `getTestFile` to optionally take in a path prefix string which is used to dynamically update the expected file path in the assertion, so that we can test classic vs addon build file paths.


packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,14 +56,40 @@ function getProjectConfiguration(project) {
...
+function _addonFound(pathSegments, projectConfiguration) {
+  const addonPaths = projectConfiguration.pkg['ember-addon'].paths;
+
+  if (!addonPaths) {
+    return;
+  }
+
+  const addonNames = addonPaths.map((path) => path.replace('lib/', ''));
+
+  return pathSegments.filter((addonName) => addonNames.includes(addonName))[0];
```
scalvert, line 68: Should we be using filter here? We're trying to find the addon name from the pathSegments, right? Should we not instead use `find`?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,14 +56,40 @@ function getProjectConfiguration(project) {
...
+}
+
+function _getParsedAddonPathSegments(pathSegments, addonName) {
+  const RELATIVE_ADDON_PATH_ROOT = 1;
+
+  pathSegments[0] = 'lib';
```
scalvert, line 74: I'm a bit confused as to what's going on here. We're mutating the passed in array, and replacing it's 0th index. What is the rationale here?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -90,39 +85,86 @@ module.exports = function (defaults) {
...
-        tests: {
-          unit: getTestFiles('with-hooks-assert-includes-test.js'),
+  const lib = {};
+
+  names.forEach((name) => {
+    project.pkg['ember-addon'].paths.push(`lib/${name}`);
+    lib[name] = {
+      'package.json': `{
+        "name": "${name}",
+        "version": "${version}",
```
scalvert, line 109: Why is this change necessary?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -90,39 +85,86 @@ module.exports = function (defaults) {
...
+        "keywords": ["ember-addon"]
+      }`,
+      'index.js': `module.exports = {
+        name: '${name}',
```
scalvert, line 113: Why is this change necessary?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -90,39 +85,86 @@ module.exports = function (defaults) {
...
   });
 }
 
+function getAddonTestFile(file, addonName) {
+  const result = {};
+  const tmpFile = readFileSync(join(__dirname, '__fixtures__', file), { encoding: 'utf-8' });
+
+  result[file] = tmpFile.replace('addon-name-placeholder', addonName);
+  return result;
+}
```
scalvert, line 166: You could probably just extend the other test helper that gets test files to optionally do this replacement if a param is passed in rather than having to build a whole new function to do it.

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -155,6 +197,7 @@ Scenarios.fromProject(baseApp)
   })
   .forEachScenario((scenario) => {
     describe(scenario.name, () => {
+      const numberOfTests = scenario.name.includes('classicInRepoAddon-app') ? '7' : '5';
```
scalvert, line 200: I'd probably try to avoid doing this type of thing vs. passing in extra data into the tests. If you do this, it's going to quickly become a mess if you're having to support multiple variations. The tests will become a conditional hellscape, and will be extremely hard to read. Instead, you should determine if there's a way to pass additional data to the scenario by attaching that data from inside the scenario itself.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/utils.js
glnster: We want to change the filepath we get: `tests/fake-addon/unit/with-hooks-in-repo-addon-test.js`
into this: `lib/fake-addon/tests/unit/with-hooks-in-repo-addon-test.js`.

So we're changing the 0th pathSegment to 'lib' and inserting 'tests' right after the addon name.

But I'm looking into the node `path` API to see if it can replace all this array manipulation.

packages/test-scenarios/__tests__/scenario-test.js
glnster: This was so we can have the quotes that json expects. Otherwise we get a build error about malformed json.

packages/test-scenarios/__tests__/scenario-test.js
glnster: IMO `${name}` is easier to read and understand vs `require("./package").name`. Is there a reason behind using `require("./package").name`? Maybe I'm unfamiliar and missing something, why we should use 'require' here.

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: I think this is the concern I have with using this approach (arrays for path manipulation), is that it feels unstructured and potentially error prone. None of the elements of an array have any semantic meaning, so it's really hard to review code like this. The 0th element of an array means nothing, and the fact that we have to call out the 1st index using a screaming caps constant is pretty telling here.

packages/test-scenarios/__tests__/scenario-test.js
scalvert: Oh, we should be `JSON.stringfy`ing this instead, not adding template strings with quotes.

packages/test-scenarios/__tests__/scenario-test.js
scalvert: It's the [default in the addon's blueprint](https://github.com/ember-cli/ember-addon-output/blob/master/index.js#L4). We're trying to adhere to it as closely as possible, which is the entire point. We're attempting to recreate as close as possible the _actual_ structure that we want to test. If you divert from that and make up your own structure, it's not testing as close to the real thing as possible.

glnster: Ah! Thank you. Yes, we want to adhere to as-close-as-possible actual structure. Reverted to use require. Thanks.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I'm currently refactoring and using `path` API.

glnster: My latest changes on Friday have refactored this to use `path` API instead of arrays.

packages/test-scenarios/__tests__/scenario-test.js
glnster: I'll work on figuring this out. We expect slightly different file path results for classic vs in-repo. And for in-repo scenarios, we have a few more tests because we're testing for multiple addons case, e.g. 13 tests vs 5.

scalvert: So maybe we just need to split things out into multiple scenario iterations?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -63,80 +59,17 @@ describe('getNormalizedFilePath', () => {
...
-  const projectConfiguration = {
-    pkg: {
-      name: 'test-app',
-      'ember-addon': {},
-    },
-  };
-
-  it('returns stripped file path as expected', () => {
-    expect(getNormalizedFilePath(fileOpts.embroiderBuildPath, projectConfiguration)).toBe(
+    }, projectConfiguration)).toBe(
```
scalvert, line 62: Let's avoid using `toBe` here, as it's not intuitive what you're expecting to assert here. I know we have string -> string comparison, but in that case I think `toEqual` makes much more sense (and reads better).

Reference: https://dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+  const embroiderRegex = new RegExp(`embroider(${path.sep})(.{6})`);
+  const embroiderPrefix = filename.match(embroiderRegex);
+  const addonFound = _addonFound(filename, projectConfiguration);
+
+  if (embroiderPrefix) {
+    const embroiderPrefixIndex = filename.search(embroiderRegex);
+    let embroiderPath = filename.substring(embroiderPrefixIndex);
+
+    if (filename.includes(rootDirWithBase)) {
+      embroiderPath = embroiderPath.replace(rootDirWithBase, '')
```
scalvert, line 90: Can we use `path.relative` here instead?

eg.

```js
path.relative('some/base', 'some/base/tests/acceptance/my-test.js') // -> 'tests/acceptance/my-test.js'
```

In general, we should always favor specific APIs for manipulation of semi-structured data over direct manipulation.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+  }
+
+  if (addonFound) {
+    const rootDirBaseWithProjectPrefix = path.join(rootDirWithBase, projectPathPrefix);
+    const sourceAddonPath = path.join(rootDirBaseWithProjectPrefix, 'tests', addonFound);
+    const targetAddonPath = path.join(rootDirBaseWithProjectPrefix, 'lib', addonFound, 'tests');
+    const parsedAddonPath = filename.replace(sourceAddonPath, targetAddonPath);
 
-    filename = isEmbroider ? _getParsedEmbroiderFilepath(pathSegments) : _getParsedClassicFilepath(pathSegments, projectConfiguration);
+    return parsedAddonPath.replace(rootDirBaseWithProjectPrefix + path.sep, '');
```
scalvert, line 103: `path.relative`?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+  const addonFound = _addonFound(filename, projectConfiguration);
+
+  if (embroiderPrefix) {
+    const embroiderPrefixIndex = filename.search(embroiderRegex);
+    let embroiderPath = filename.substring(embroiderPrefixIndex);
+
+    if (filename.includes(rootDirWithBase)) {
+      embroiderPath = embroiderPath.replace(rootDirWithBase, '')
+    }
+    embroiderPath = embroiderPath.replace(embroiderPrefix[0] + path.sep, '');
```
scalvert, line 92: `path.relative`?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -26,23 +24,22 @@ module.exports = function (defaults) {
...
+        'with-multiple-modules-test.js',
+      ]),
     },
   });
 }
 
 async function classicInRepoAddon(project) {
   await classic(project);
-  await addInRepoAddon(project, 'fake-addon');
+  await addInRepoAddon(project, ['fake-addon', 'fake-addon-two']);
```
scalvert, line 42: Rather than making this take multiple names, I would probably have called this twice in a row. What happens if we want different version numbers for each addon?

packages/test-scenarios/__tests__/scenario-test.js
glnster: Addressed by passing test info from each scenario and merging just those test files.

packages/test-scenarios/__tests__/scenario-test.js
glnster: I separated this out to another scenario test and calling addInRepoAddon twice.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I'll address this in a follow up PR.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I'll address these embroider path handling in a follow up PR.

packages/babel-plugin-ember-test-metadata/utils.js
glnster: I'll address these embroider path handling in a follow up PR.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
-function getNormalizedFilePath(fileOpts, projectConfiguration) {
-  let { root, filename } = fileOpts;
-  const pathSegments = filename.split(path.sep);
-  const isEmbroider = pathSegments.includes('embroider');
+function getNormalizedFilePath({ root, filename }, projectConfiguration) {
+  const projectPathPrefix = projectConfiguration.pkg.name;
+  const rootDirWithBase = path.join(path.parse(root).dir, path.parse(root).base);
+  const embroiderRegex = new RegExp(`embroider(${path.sep})(.{6})`);
```
scalvert, line 81: Can we move this to be a module-scoped constant so we don't have to create a new RegExp for every iteration? I think this pattern should be static for the entire build, correct?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+  const embroiderPrefix = filename.match(embroiderRegex);
```
scalvert, line 82: Is this actually `embroiderSegment`? As we're looking for the presence of `embroider` as part of the path, no?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
-  const projectNamePathSeparators = projectConfiguration.pkg.name.split(path.sep);
-
-  pathSegments.splice(
-    0,
-    pathSegments.indexOf(projectNamePathSeparators[0]) + projectNamePathSeparators.length
-  );
-
-  return pathSegments.join(path.sep);
-}
+function _addonFound(filename, projectConfiguration) {
```
scalvert, line 59: Looking at how this is used in `getNormalizedFilePath` below, it looks like this should be named `getAddonName`, as it returns the name of the addon found in the path. This makes it make a bit more sense when it's passed as a param into the `path.join` below, as otherwise it seems like this is returning a boolean since the return value of this currently is `addonFound`.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
-function getNormalizedFilePath(fileOpts, projectConfiguration) {
-  let { root, filename } = fileOpts;
-  const pathSegments = filename.split(path.sep);
-  const isEmbroider = pathSegments.includes('embroider');
+function getNormalizedFilePath({ root, filename }, projectConfiguration) {
+  const projectPathPrefix = projectConfiguration.pkg.name;
+  const rootDirWithBase = path.join(path.parse(root).dir, path.parse(root).base);
+  const embroiderRegex = new RegExp(`embroider(${path.sep})(.{6})`);
+  const embroiderPrefix = filename.match(embroiderRegex);
+  const addonFound = _addonFound(filename, projectConfiguration);
```
scalvert, line 83: Based on my comment above, I think this variable should be named `addonName`. This will make it read more logically below when passed to `path.join`.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+      embroiderPath = embroiderPath.replace(rootDirWithBase, '')
+    }
+    embroiderPath = embroiderPath.replace(embroiderPrefix[0] + path.sep, '');
+
+    return embroiderPath;
+  }
+
+  if (addonFound) {
+    const rootDirBaseWithProjectPrefix = path.join(rootDirWithBase, projectPathPrefix);
+    const sourceAddonPath = path.join(rootDirBaseWithProjectPrefix, 'tests', addonFound);
```
scalvert, line 99: Trying to understand what we're trying to model here. Could you help me understand what Ember app directory structure we expect this is representing?

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,40 +56,56 @@ function getProjectConfiguration(project) {
...
+    const targetAddonPath = path.join(rootDirBaseWithProjectPrefix, 'lib', addonFound, 'tests');
```
scalvert, line 100: I've mentioned this before, but this path only models an app with in-repo addons that _also_ use `ember-add-in-repo-tests`. So long as this is what we're trying to represent here, this makes sense.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/utils.js
glnster: Here's what they model. I think if I add comments, it will be helpful.

```
    const rootDirBaseWithProjectPrefix = path.join(rootDirWithBase, projectPathPrefix);
    //E.g. /private/var/folders/2j/some-hash/T/tmp-hash/@my-package/app-template

    const sourceAddonPath = path.join(rootDirBaseWithProjectPrefix, 'tests', addonName);
    // E.g. /private/var/folders/2j/some-hash/T/tmp-hash/@my-package/app-template/tests/my-addon

    const targetAddonPath = path.join(rootDirBaseWithProjectPrefix, 'lib', addonName, 'tests');
    // E.g. /private/var/folders/2j/some-hash/T/tmp-hash/@my-package/app-template/lib/my-addon/tests
```

scalvert: Right, but what do they represent in terms of actual, real-world apps. Can you point to any that show this structure?

For instance, it's not clear to me what the first represents. Is it an app with a scope? Why would it have a scope?

For the second, what structure ever has addons nested under a tests directory in the app?

For the third, this represents an approximation of what an app that's using `ember-add-in-repo-tests` would look like, correct?

packages/babel-plugin-ember-test-metadata/utils.js
glnster: Yes, this is what we're looking for and modeling. Referring to the structure here https://github.com/sangm/ember-add-in-repo-tests

PR #42, opened 2021-09-23
Add util test cases for different path scenarios

ckundo:
Adds new utils for each of the path normalization scenarios. Adds tests for permutations of those paths. Updates `getNormalizedFilePath` to handle all variations. Updates scenario tests.


packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +30,112 @@ describe('getNodeProperty', () => {
...
+  const config = {
+    pkg: {
+      name: appName
+    },
+  }
+
+  describe("given file path from a classic build", () => {
+    it("returns the normalized filepath", () => {
+      const expectedPath = path.join('tests', 'acceptance', 'example-test.js');
+      const preFormattedPath = path.resolve(appName, moduleName, expectedPath);
```
scalvert, line 46: I don't think we want `path.resolve` here. I think we should try to mirror, exactly, the values we identified, and each test should test multiple variations of paths.

We should just add each of these in our tests.

```
/Users/tester/workspace/personal/test-bed/classic/classic/tests/acceptance/foo-test.js
/Users/tester/workspace/personal/test-bed/classic/classic/tests/acceptance/subdir/foo-test.js
/Users/tester/workspace/personal/test-bed/classic/classic/tests/acceptance/subdir/subdir/foo-test.js
```

I've also thought about whether we need to use `path.join` or not, and I don't think we do. We don't need to support windows.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +30,112 @@ describe('getNodeProperty', () => {
...
+  const moduleName = "example-app";
+  const addonName = "example-addon";
+  const addonPath = path.join('lib', addonName);
+  const config = {
+    pkg: {
+      name: appName
+    },
+  }
+
+  describe("given file path from a classic build", () => {
```
scalvert, line 43: I don't think you need the describes, but just a single test for each of the scenarios. If we'd rather have separate utils for each of the normalizers, that's fine too.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -85,13 +85,54 @@ function _getParsedEmbroiderFilepath(pathSegments) {
...
+      pathSegments.splice(0, pathSegments.lastIndexOf(addonName) + 1);
+      return ['lib', addonName, "tests", ...pathSegments].join(path.sep)
+    } else {
+      const projectNamePathSeparators = projectConfiguration.pkg.name.split(path.sep);
+      pathSegments.splice(
+        0,
+        pathSegments.lastIndexOf(root) + projectNamePathSeparators.length
+      );
+      return pathSegments.join(path.sep);
+    }
```
scalvert, line 116: I think each of these should be separate utils, or normalizers. That way we can compose the normalizers based on what we need to support.

packages/babel-plugin-ember-test-metadata/__tests__/__snapshots__/index-test.js.snap
```diff
@@ -23,7 +23,7 @@ import { click, getTestMetadata } from '@ember/test-helpers';
 module('Acceptance | example test embroider prefix', function (hooks) {
   hooks.beforeEach(function () {
     let testMetadata = getTestMetadata(QUnit.config.current.testEnvironment);
-    testMetadata.filePath = 'pathtest/embroider-prefix-test.js';
```
ckundo, line 26: i don't think this case should ever exist.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
   return pathSegments.join(path.sep);
 }
 
+function _getEmbroiderPathWithAddon(pathSegments, addons) {
+  const addonNames = addons.paths.map(addonPath => addonPath.split(path.sep)[1]);
+  const addonName = addonNames.find(name => pathSegments.includes(name));
+
+  if (pathSegments.includes(addonName)) {
+    pathSegments.splice(0, pathSegments.lastIndexOf(addonName) + 1);
+    const hasTestsPrefix = pathSegments[0] === "tests";
```
ckundo, line 91: this is a workaround for scenario tests including duplicate `tests` prefix.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
+  if (pathSegments.includes(addonName)) {
+    pathSegments.splice(0, pathSegments.indexOf("ember-add-in-repo-tests") + 1);
+    return [...pathSegments].join(path.sep)
+  } else {
+    return _getClassicPath(preformattedPath, projectName);
+  }
+}
 
+function _getEmbroiderPath(pathSegments) {
+  pathSegments.splice(0, pathSegments.lastIndexOf('tests'));
```
ckundo, line 81: I _think_ this is safe to look for `tests` instead of `embroider`, and this approach fixes some scenario testing config issues.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
 
-  it('returns unmodified file path when path does not include "embroider" as a segment', () => {
-    expect(getNormalizedFilePath(fileOpts.normalFilePath, projectConfiguration)).toBe(
-      path.join('this', 'is', 'not-an-embroider', 'path')
-    );
-  });
-});
+  describe("given file path from a classic build", () => {
+    beforeEach(() => {
+      appRoot = "/Users/tester/workspace/personal/test-bed/classic";
```
scalvert, line 36: This is just a const, and can be lifted up to function scope and declared as const using screaming caps.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
 
-describe('_getParsedClassicFilepath', () => {
-  it('returns file path from classic build correctly', () => {
-    expect(
-      _getParsedClassicFilepath(['test-app', 'tests', 'acceptance', 'my-test.js'], {
+    it("returns the normalized filepath", () => {
+      const expectedPath = "tests/acceptance/foo-test.js";
+      const preFormattedPath = "/Users/tester/workspace/personal/test-bed/classic/classic/tests/acceptance/foo-test.js";
```
scalvert, line 41: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const opts = { filename: preFormattedPath, root: "/Users/tester/workspace/personal/test-bed/classic" };
```
scalvert, line 42: I suspect we could have a simple test util that just merges things together, as this is used in multiple places.

```js
const DEFAULT_OPTS = { root: '/Users/tester/workspace/personal/test-bed/classic' };

function getOpts(filename) {
  return Object.assign({}, DEFAULT_OPTS, { filename });
}
```

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
-            'ember-addon': {},
-          },
-        }
-      )
-    ).toBe(path.join('tests', 'acceptance', 'my-test.js'));
+      }
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 51: Let's not use `toBe`. We should use `toEqual` instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. https://dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
+
+    it("returns the normalized filepath for the in app test", () => {
+      const expectedPath = "tests/acceptance/foo-test.js";
+      const preFormattedPath = "/Users/tester/workspace/personal/test-bed/classic/classic/tests/acceptance/foo-test.js";
```
scalvert, line 56: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const opts = { filename: preFormattedPath, root: "/Users/tester/workspace/personal/test-bed/classic" };
```
scalvert, line 57: Swap to util above.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+          name: "classic"
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ['lib/bar']
+      }
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 69: Let's not use toBe. We should use toEqual instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
+
+    it("returns the normalized filepath with the addon name", () => {
+      const expectedPath = `lib/bar/tests/acceptance/foo-test.js`
+      const preFormattedPath = `${appRoot}/tests/ember-add-in-repo-tests/lib/bar/tests/acceptance/foo-test.js`;
```
scalvert, line 74: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const opts = { filename: preFormattedPath, root: appRoot };
```
scalvert, line 75: Swap to util above.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+          name: "classic"
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ['lib/bar']
+      }
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 87: Let's not use toBe. We should use toEqual instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
-  );
-  const pathSegments = filename.split(path.sep);
-
-  it('returns file path from classic build correctly', () => {
-    expect(_getParsedEmbroiderFilepath(pathSegments)).toBe(
-      path.join('tests', 'acceptance', 'my-test.js')
-    );
+  describe("given file path from an embroider build", () => {
+    beforeEach(() => {
+      appRoot = '/private/var/folders/abcdefg1234/T/embroider/098765';
```
scalvert, line 93: This is just a const, and can be lifted up to function scope and declared as const using screaming caps.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
+
+    it("returns the normalized filepath for app tests", () => {
+      const expectedPath = `tests/acceptance/foo-test.js`
+      const preFormattedPath = `${appRoot}/tests/acceptance/foo-test.js`;
```
scalvert, line 98: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const config = {
+        pkg: {
+          name: "example-app",
+        },
+      }
+      const opts = { filename: preFormattedPath, root: appRoot };
```
scalvert, line 104: Swap to util above.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 108: Let's not use toBe. We should use toEqual instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
+
+    it("returns the normalized filepath for addon app tests", () => {
+      const expectedPath = `tests/acceptance/foo-test.js`
+      const preFormattedPath = `${appRoot}/tests/acceptance/foo-test.js`;
```
scalvert, line 113: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const config = {
+        pkg: {
+          name: "example-app",
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ["lib/bar"]
+      }
+      const opts = { filename: preFormattedPath, root: appRoot };
```
scalvert, line 122: Swap to util above.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 126: Let's not use toBe. We should use toEqual instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+    });
+
+    it("returns the normalized filepath with the addon name for addon tests", () => {
+      const expectedPath = `lib/bar/tests/acceptance/foo-test.js`
+      const preFormattedPath = `${appRoot}/tests/ember-add-in-repo-tests/lib/bar/acceptance/foo-test.js`;
```
scalvert, line 131: My brain always wants the inputs listed before the outputs as it makes me have to think more with the reverse. It's a nit but it helps with readability. Can we reverse these?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const config = {
+        pkg: {
+          name: "example-app",
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ["lib/bar"]
+      }
+      const opts = { filename: preFormattedPath, root: appRoot };
```
scalvert, line 140: Swap to util above.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+
+      const formattedPath = getNormalizedFilePath(opts, config);
+
+      expect(formattedPath).toBe(expectedPath);
```
scalvert, line 144: Let's not use toBe. We should use toEqual instead. While this doesn't really matter as much with strings, it's a good habit to be in and it's more explicit than a fuzzy assertion. dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+        pkg: {
+          name: "example-app",
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ["lib/bar"]
+      }
+      const opts = { filename: preFormattedPath, root: appRoot };
+
+      const formattedPath = getNormalizedFilePath(opts, config);
```
scalvert, line 142: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`?

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
```
scalvert, line 124: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const expectedPath = `tests/acceptance/foo-test.js`
+      const preFormattedPath = `${appRoot}/tests/acceptance/foo-test.js`;
+      const config = {
+        pkg: {
+          name: "example-app",
+        },
+      }
+      const opts = { filename: preFormattedPath, root: appRoot };
+
+      const formattedPath = getNormalizedFilePath(opts, config);
```
scalvert, line 106: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
+      const config = {
+        pkg: {
+          name: "classic"
+        },
+      }
+      config.pkg['ember-addon'] = {
+        paths: ['lib/bar']
+      }
+
+      const formattedPath = getNormalizedFilePath(opts, config);
```
scalvert, line 85: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
```
scalvert, line 67: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
...
-          pkg: {
-            name: `@parent-repo-name${path.sep}test-app`,
-            'ember-addon': {},
-          },
-        }
-      )
-    ).toBe(path.join('tests', 'acceptance', 'my-test.js'));
+      }
+
+      const formattedPath = getNormalizedFilePath(opts, config);
```
scalvert, line 49: Can we name this `normalizedFilePath`, since you're calling a function called `getNormalizedFilePath`, which is intended to return a `normalizedFilePath`.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
+    .slice(preformattedPath.lastIndexOf(projectName), preformattedPath.length)
+    .replace(`${projectName}/`, '');
 
-  pathSegments.splice(
-    0,
-    pathSegments.indexOf(projectNamePathSeparators[0]) + projectNamePathSeparators.length
-  );
-
-  return pathSegments.join(path.sep);
+  return formattedPath;
```
scalvert, line 64: We should seriously consider not using arrays for path manipulation. I made this comment before, and I believe Glenn made those changes to use `path.parse` instead. Using arrays is error prone when you're working with structured strings, so we should do our best to avoid things that could get us into trouble in the future.

We don't have to do those changes in this PR.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
   return parsedProjectConfiguration;
 }
 
-function _getParsedClassicFilepath(pathSegments, projectConfiguration) {
-  const projectNamePathSeparators = projectConfiguration.pkg.name.split(path.sep);
+function _getClassicPath(preformattedPath, projectName) {
```
scalvert, line 59: Let's order these params logically from their scope of what they represent, wider scope (project) to narrow scope (filename). I also think we can make the naming a bit easier to understand.

```suggestion
function _getRelativePathForClassic(projectName, filePath) {
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
-    pathSegments.indexOf(projectNamePathSeparators[0]) + projectNamePathSeparators.length
-  );
-
-  return pathSegments.join(path.sep);
+  return formattedPath;
 }
 
-function _getParsedEmbroiderFilepath(pathSegments) {
-  const RELATIVE_PATH_ROOT = 2;
+function _getClassicPathWithAddon(preformattedPath, projectName, addons) {
```
scalvert, line 67: The relevant part here is that it's an in-repo addon path. Let's name it such.

```suggestion
function _getRelativePathForClassicInRepo(projectName, addons, filePath) {
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
-  pathSegments.splice(0, pathSegments.lastIndexOf('embroider') + RELATIVE_PATH_ROOT);
+  if (pathSegments.includes(addonName)) {
+    pathSegments.splice(0, pathSegments.indexOf("ember-add-in-repo-tests") + 1);
+    return [...pathSegments].join(path.sep)
+  } else {
+    return _getClassicPath(preformattedPath, projectName);
+  }
+}
 
+function _getEmbroiderPath(pathSegments) {
```
scalvert, line 80: This signature differs from those above. We should try to keep them consistent, so let's pass the `filePath` in here as well and do the split in this function.

```suggestion
function _getRelativePathForEmbroider(filePath) {
```

packages/babel-plugin-ember-test-metadata/utils.js
scalvert: I don't think so. What if there's nested tests directories? I know that's odd, but I would think that much more likely than having multiple `embroider` directories. We should stick with using `embroider`.

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
+  pathSegments.splice(0, pathSegments.lastIndexOf('tests'));
   return pathSegments.join(path.sep);
 }
 
+function _getEmbroiderPathWithAddon(pathSegments, addons) {
```
scalvert, line 85:
```suggestion
function _getRelativePathForEmbroiderInRepo(filePath, addons) {
```

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
+  let { filename } = fileOpts;
   const pathSegments = filename.split(path.sep);
   const isEmbroider = pathSegments.includes('embroider');
-
-    filename = isEmbroider ? _getParsedEmbroiderFilepath(pathSegments) : _getParsedClassicFilepath(pathSegments, projectConfiguration);
-
-  return path.relative(root, filename);
+  const addons = projectConfiguration.pkg["ember-addon"];
+  const hasAddons = addons && addons.paths;
+  const projectName = projectConfiguration.pkg.name;
```
scalvert, line 111: I don't feel as strongly about this, but it may be nice to pass the `projectConfiguration` in to the utils below instead of the individual props. This will result in a simplified signature. 

This doesn't apply to hasAddons (more on that below).

packages/babel-plugin-ember-test-metadata/utils.js
```diff
@@ -56,48 +56,73 @@ function getProjectConfiguration(project) {
...
+
+  if (!isEmbroider && !hasAddons) {
```
scalvert, line 113: I don't think this logic is correct. It's not about whether the project has addons, but whether the `filename` is from an addon path. The logic here is we need to determine if the app is using in-repo addons, then further testing if the path is either the classic case or the classic in-repo addon case, so this should actually be a nested conditional.

The same applies to those below.

packages/test-scenarios/__tests__/__fixtures__/with-hooks-assert-includes-test.js
```diff
@@ -7,8 +7,8 @@ module('Acceptance | with-hooks-test', function (hooks) {
   });
 
   test('example', async function (assert) {
-    assert.ok(
-      getTestMetadata(this).filePath.includes('tests/unit/with-hooks-assert-includes-test.js')
-    );
+    const matcher = /^tests\/unit\/with-hooks-assert-includes-test\.js$/g;
```
scalvert, line 10: Why do we need a regex here?

packages/test-scenarios/__tests__/scenario-test.js
```diff
@@ -6,12 +6,19 @@ const { merge } = require('lodash');
 jest.setTimeout(500000);
 
 async function classic(project) {
+  project.pkg.scenarioTesterTests = {
```
scalvert, line 9: I'll just make one general comment for all these scenarios, and I'm sure you're aware, but they're not complete. They need to match the specific variations we identified as cases we need to support.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
```diff
@@ -32,111 +29,120 @@ describe('getNodeProperty', () => {
 });
 
 describe('getNormalizedFilePath', () => {
```
scalvert, line 31: General comment here, but we're missing cases for _all_ the paths we need to support. We should have tests for each of the new utils created, so we've covered all test cases.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
ckundo: I get the intent, but I prefer being procedural about tests for the sake of clarity.

packages/test-scenarios/__tests__/scenario-test.js
scalvert: Just re-read your comment above about adding follow ons. I'd do that but write tests for the internal utils.

packages/babel-plugin-ember-test-metadata/__tests__/utils-test.js
scalvert: Just re-read your comment above about adding follow ons. I'd do that but write tests for the internal utils.

scalvert (approved):

PR #44, opened 2021-09-24
Adds config for package name and if running under embroider

scalvert:
Adds config to allow for identifying when running under Embroider.

```js
'use strict';

const EmberApp = require('ember-cli/lib/broccoli/ember-app');

module.exports = function (defaults) {
  let app = new EmberApp(defaults, {
    babel: {
      plugins: [
        [
          require.resolve('babel-plugin-ember-test-metadata'),
          {
            enabled: !!process.env.BABEL_TEST_METADATA,
            packageName: defaults.project.pkg.name,
            isUsingEmbroider: true,
          },
        ],
      ],
    },
  });

  // additional configuration

  return app.toTree();
};
```

When using Embroider, you would explicitly pass `true` to the plugin.


ckundo (approved):
lgtm

PR #35, opened 2021-09-08
Adding test-scenarios package for acceptance-style tests

scalvert:
Adds structure for test scenarios using `scenario-tester`.

Currently adds:
- classic tests (passing, using inadequate assertions which only match part of `filePath`)
- embroider tests (passing, using inadequate assertions which only match part of `filePath`)

Recommendation:
- Merge this PR once TODOs are completed
- Branch, and update the tests to match on the exact expected path for all tests (eg. `tests/unit/some-test.js`)
- Fix issue with both classic and embroider builds, using failing tests as a guide.

TODO (in a follow up PR):

- Add scenarios for in-repo engine and addon

PR #127, opened 2022-03-12
Switch from absolute to relative path for babel file

ckundo:
- use sourceFilePath (relative path) instead of full filepath.
- use regexes to simplify path parsing and replacement.
- move implementation to lib/
- consolidate utilities into one file now that they are very simple.
- update implementation and tests to support yarn workspaces.
- remove babel plugin tester so we can more easily stub the babel filepath in tests.


packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -1,36 +1,32 @@
...
  * @param {object} opts Babel state.file.opts which include root and filename props
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
-function getNormalizedFilePath({ packageName, isUsingEmbroider, filename, root }) {
-  if (!isUsingEmbroider) {
-    if (filename.includes('ember-add-in-repo-tests')) {
-      return _getRelativePathForClassicInRepo(filename);
-    }
+function getNormalizedFilePath({ packageName, filename, isUsingEmbroider, root, sourceFileName }) {
+  let file = sourceFileName || filename;
```
ckundo, line 7: there a step in some ember build processes that doubled the package path, e.g. `/path/to/project/extended/extended/tests/my-test.js`. `sourceFileName` returns `extended/tests/my-test.js` which is correct.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -1,36 +1,32 @@
...
+    const regex = new RegExp(`^.*embroider/.{6}/`);
+    file = file.replace(regex, '');
   }
+
+  if (file.includes('ember-add-in-repo-tests')) {
+    const regex = new RegExp(`^.*ember-add-in-repo-tests/`);
+    file = file.replace(regex, '');
+  }
+
+  return file;
```
scalvert, line 29: I think this results in code that is harder to read. The reason we split the different permutations out into well factored units was so that we could encapsulate the logic into semantically named functions, which makes it much easier to go back through and understand the intent of each. Moving them all into inline ifs defeats that purpose.

If I'm not mistaken, this was an early direction that was pursued only to be discarded in favor of the well factored units into functions approach.

Additionally, it made testing these permutations discrete, and readable. We've also removed a lot of the tests that had provided us value through quick iteration.

I'm not in favor of us continuing down this path.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -1,36 +1,32 @@
...
-  _getRelativePathForClassicInRepo,
-  _getRelativePathForEmbroider,
-  _getRelativePathForEmbroiderInRepo,
-} = require('./get-relative-paths');
-
-const path = require('path');
-
 /**
  * Get a normalized file path, based on whether the app build is classic or with Embroider
  * @param {object} opts Babel state.file.opts which include root and filename props
```
scalvert, line 3: Can we use jsdoc to list out the actual Babel `state.file.opts`? This will make it easier for readers to understand where these params are originating from, and give some clue into their purpose.

scalvert:
Part of the challenge with limiting the scope of our changes to the units that manage manipulating file paths is that we're likely to miss regressions from the underlying system that we actually integrate with. More specifically, we previously encountered "drift" from the file path parsing code and the permutations of the real Ember apps under test. It was only once we employed [scenario tester](https://github.com/babel-plugin-ember-test-metadata/babel-plugin-ember-test-metadata/pull/35) that we were able to not only resolve all our outstanding issues and support the different configurations of apps that we do, but it allowed us to guard against future regressions due to changes in the underlying framework.

I think we need to do the following:

- Create a separate test app that is structured how a conventional `yarn workspaces` repository would be structured
- Create a separate number of scenarios that test against that real workspaces setup. This can and should include conventional `yarn workspaces` setups (`/packages/ember-app`) and those that include addons in a separate package directory (`/packages/addons` as a sibling to `/packages/ember-app`).

This will greatly increase our chances of ensuring that our plugin still works for all scenarios, _and_ will provide the added benefit of ensuring we don't regress. In addition, any further bugs can be confidently resolved knowing that we haven't regressed any of the other supported scenarios.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
ckundo: thanks for the input. Discrete functions and tests were be helpful in ironing out the implementation and building confidence in our design choices. That said, we have the same cases covered in `index-test.js`. This approach made sense to me because there's far less indirection, everything you want to know about the transformations are in one place. what are your thoughts?

ckundo: I could also split out these blocks into individual helpers in the same file, keeping the newer implementation and readding the unit tests.. hows that sound?

ckundo:
This PR is in response to a regression that was due to invalid absolute paths being sent to the plugin, unrelated to workspaces. I have a followup PR for workspaces testing.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -1,36 +1,32 @@
...
-      return _getRelativePathForEmbroiderInRepo(filename);
-    }
+  if (file.startsWith(`${packageName}/`)) {
+    const regex = new RegExp(`^${packageName}/`);
+    file = file.replace(regex, '');
+  }
 
-    return _getRelativePathForEmbroider(filename);
+  if (isUsingEmbroider) {
+    const regex = new RegExp(`^.*embroider/.{6}/`);
```
drewlee, line 20: Given that this is an OSS project, the other benefit of the previous implementation was that it accommodated Windows based file paths. If we do go w/regular expressions, we should use `path.sep` instead of hardcoding the slashes.

scalvert: I need to understand the performance implications of creating so many `RegExp` instances. Since this plugin runs in our build pipeline, we have to be very sensitive to memory or CPU pressure as a result of changes that are added to babel plugins.

Have you done any performance analysis in a large repository related to this change?

ckundo:
tbh, I did this so it was easier to understand so I could implement the fix 😂 , but yea I can revert it all and just pass the `sourceFileName`.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
ckundo: IMO the tests are effective in describing the behavior, but that's easy enough to split out utility methods for those blocks if you think it's easier to maintain and read. I'm reverting to the original design, but this code works right now.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
ckundo: not for this change; in fact, I don't know that this project has ever had a perf analysis. do you remember differently? How would you want to see this benchmarked?

scalvert: We did some analysis when we originally wrote it to ensure build times weren't going to be adversely affected. I'll have to dig it up, but I believe we used broccoli's debug flags which expose Heimdall counters to determine the overall plugin perf.

eg.

```shell
DEBUG_DEPTH=5 DEBUG_LEVEL=debug DEBUG=broccoli-merge-trees ember build
```

scalvert:
Thanks, @ckundo! I think doing these changes incrementally will be _much_ easier to review. If we want to move ahead with any design changes we can discuss those for future PRs.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -9,10 +9,17 @@ const path = require('path');
 
 /**
  * Get a normalized file path, based on whether the app build is classic or with Embroider
- * @param {object} opts Babel state.file.opts which include root and filename props
+ * @param {Object} opts Babel state.file.opts which include root and filename props
+ * @param {string} opts.packageName the name of the package specified in babel plugin options
+ * @param {boolean} opts.isUsingEmbroider flag for embroider builds as specified in babel plugin options
+ * @param {string} opts.filename absolute path the the name of the file being visited
```
drewlee, line 15: Typo: "the the"

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -9,10 +9,17 @@ const path = require('path');
...
+ * @param {string} opts.root path to the project root (defaults to cwd)
+ * @param {string} opts.sourceFileName path.basename(opts.filenameRelative) when available, or "unknown"
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
-function getNormalizedFilePath({ packageName, isUsingEmbroider, filename, root }) {
+function getNormalizedFilePath({ packageName, isUsingEmbroider, filename, root, sourceFileName }) {
+  if (sourceFileName && sourceFileName !== 'unknown') { return sourceFileName }
```
drewlee, line 21: Prefer return statements on their own line:
```
if (sourceFileName && sourceFileName !== 'unknown') {
  return sourceFileName;
}
```

drewlee (commented):
A couple of minor comments, otherwise LGTM.

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -9,15 +9,27 @@ const path = require('path');
...
     if (filename.includes('ember-add-in-repo-tests')) {
       return _getRelativePathForClassicInRepo(filename);
     }
 
+    if (sourceFileName &&
+      !path.isAbsolute(sourceFileName) &&
+      !sourceFileName.startsWith(packageName) &&
+      sourceFileName !== 'unknown') {
+      return sourceFileName;
+    }
```
scalvert, line 31: Can we make a comment to identify what specifically what case we're trying to handle here?

scalvert:
Hey @ckundo, can we make sure reviewers are re-reviewing after changes before we merge. This will make sure we're only merging code that makes sense/meets our standards and ensures the code remains clear and well factored. Thanks!

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
scalvert: Both would be preferable. People are going to go to the implementation first, and only when that implementation isn't clear will they usually resort to looking at the test cases. 

If you update the test case titles too, which would be preferred, it'd be useful to know what use case this is handling rather than articulating the implementation details.

thousand:
@scalvert since the change hasn't been released yet and and improved testing probably won't land until @ckundo is back from CSUN, does it make sense to revert this and add it back with the improved testing? Want to make sure the repo is healthy in case anything else needs to be changed in the mean time...

scalvert:
Thanks, @thousand. That did cross my mind - I think that's prudent. I'd much prefer we have a workable, well tested solution merged as opposed to one that's in progress. I'll go ahead and revert for now.

PR #129, opened 2022-03-18
Revert "Switch from absolute to relative path for babel file"

scalvert:
Reverts babel-plugin-ember-test-metadata/babel-plugin-ember-test-metadata#127


ckundo:
this revert reintroduces a bug for some of the projects that currently rely on absolute paths. will follow up directly.

PR #136, opened 2022-04-04
Add scenario test and app for yarn workspaces

ckundo:
- Add a new config optional entry for workspace root.
- The workspace root value will be used to specify the base directory for the path separation.
- We cannot simply override the `packageName` field because  broccoli sends us a path that is not composed in a way that makes that possible. For example this is an approximate value that the plugin get for a yarn workspaces project:

        /Users/tester/workspace/personal/test-bed/classic/packages/extended/extended/tests/acceptance/foo-test.js

There's no packageName value that can be passed that will work for this case.

- `packageName: 'classic/packages/extended'` will return `extended/tests/acceptance/foo-test.js` when it should return `packages/extended/tests/acceptance/foo-test.js`.
- `packageName: 'classic'` will return `packages/extended/extended/tests/acceptance/foo-test.js` which is also incorrect.


scalvert:
What about using `packages/extended`?

ckundo:
@scalvert that would result in `extended/tests/acceptance/foo-test.js` which is missing the `packages` root.

scalvert:
Got it. Just trying to understand so that we're on the same page. When given this path:

`/Users/tester/workspace/personal/test-bed/classic/packages/extended/extended/tests/acceptance/foo-test.js`

`classic` = the root of the project
`packages` = the workspace directory (obviously)
`extended`= the Ember application's name prepended to the remaining real relative path
`extended` (the 2nd!) = the Ember application's directory root

Is this correct? Also, specifically it looks like you're running the build from the root, hence why the path is containing the full root directory (without going into too much detail, this is the special application we're aware of that has this setup). Is this also accurate?

PR #233, opened 2023-02-14
feat: add custom function as a babel option to normalize test file path

zhanwang626:
Some large Ember Apps might want to only run test for some specific set of packages at one time. In this case, it will run `ember build` in a specific package dir, and that package will build isolated. 

For example, when the test is running, the test file path will be something like:
`/Users/tester/workspace/personal/test-bed/example-app/packages/addons/example-addon/example-app/tests/acceptance/foo-test.js` 

And the expected relative path will be
`packages/addons/example-addon/tests/acceptance/foo-test.js`

To solve this problem, we can pass a custom normalized file path function as a babel option.

### DIFF
--- a/packages/babel-plugin-ember-test-metadata/README.md
+++ b/packages/babel-plugin-ember-test-metadata/README.md
@@ -120,7 +120,7 @@ You can also pass a custom function to normalize the test file path.
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function (defaults) {
-  function customNormallizedFilePath(filePath, packageName, projectRoot) {
+  function customNormalizedFilePath(filePath, packageName, projectRoot) {
     // Custom normalization
   }
 
--- a/packages/babel-plugin-ember-test-metadata/README.md
+++ b/packages/babel-plugin-ember-test-metadata/README.md
@@ -113,7 +113,7 @@ module.exports = function (defaults) {
 };
 ```
 
-You can also pass a custom function to normalize your test file path.
+You can also pass a custom function to normalize the test file path.
 ```js
 'use strict';
 
@@ -121,7 +121,7 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function (defaults) {
   function customNormallizedFilePath(filePath, packageName, projectRoot) {
-    // Your custom normalize process
+    // Custom normalization
   }
 
   let app = new EmberApp(defaults, {
@@ -133,7 +133,7 @@ module.exports = function (defaults) {
             enabled: !!process.env.BABEL_TEST_METADATA,
             packageName: defaults.project.pkg.name,
             packageRoot: '../..',
-            getCustomNormalizedFilePath: customNormallizedFilePath,
+            getCustomNormalizedFilePath: customNormalizedFilePath,
           },
         ],
       ],
--- a/packages/babel-plugin-ember-test-metadata/README.md
+++ b/packages/babel-plugin-ember-test-metadata/README.md
@@ -112,3 +112,36 @@ module.exports = function (defaults) {
   return app.toTree();
 };
 ```
+
+You can also pass a custom function to normalize your test file path.
+```js
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function (defaults) {
+  function customNormallizedFilePath(filePath, packageName, projectRoot) {
+    // Your custom normalize process
+  }
+
+  let app = new EmberApp(defaults, {
+    babel: {
+      plugins: [
+        [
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -11,13 +11,17 @@ const {
...
  * @param {string} opts.filename the absolute perceived path of the file being visited
  * @param {string} opts.root the absolute root project path as seen on disk
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
-function getNormalizedFilePath({ packageName, isUsingEmbroider, projectRoot, filename, root }) {
-  if (!isUsingEmbroider) {
+function getNormalizedFilePath({ packageName, getCustomNormalizedFilePath, isUsingEmbroider, projectRoot, filename, root }) {
+  if (typeof getCustomNormalizedFilePath === 'function') {
+
+    return getCustomNormalizedFilePath(filename, packageName, projectRoot);
```
drewlee, line 23: I'd err on the side of caution and just pass through all the parameters, as you never know what other use cases we'll discover in the near future.

packages/babel-plugin-ember-test-metadata/__tests__/get-normalized-file-path-test.js
```diff
@@ -158,4 +159,41 @@ describe('getNormalizedFilePath', () => {
...
+      
+        if (!projectRoot) {
+          return testFilePath;
+        }
+
+        const addonRoot = pathSegments.slice(0, pathSegments.lastIndexOf(packageName));
+        const projectBaseDir = basename(resolve(addonRoot.join(sep), projectRoot));
+        const projectPath = addonRoot.slice(addonRoot.indexOf(projectBaseDir) + 1).join(sep);
+        
+        return join(projectPath, testFilePath);
```
drewlee, line 180: I don't think we need all this superfluous stuff for the specific use case we're trying to support. Meaning, this can be simplified.

scalvert (approved):
This looks great! As long as @drewlee's comments are addressed this looks GTG! Thanks!

packages/babel-plugin-ember-test-metadata/README.md
```diff
@@ -111,4 +111,37 @@ module.exports = function (defaults) {
...
+```
+
+You can also pass a custom function to normalize your test file path.
+```js
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function (defaults) {
+  function customNormallizedFilePath(filePath, packageName, projectRoot) {
```
drewlee, line 123: Typo in `customNormallizedFilePath`.
customNormallizedFilePath -> customNormalizedFilePath

packages/babel-plugin-ember-test-metadata/README.md
```diff
@@ -111,4 +111,37 @@ module.exports = function (defaults) {
 
   return app.toTree();
 };
+```
+
+You can also pass a custom function to normalize your test file path.
```
drewlee, line 116: Nit:
```
You can also pass a custom function to normalize the test file path.
```

packages/babel-plugin-ember-test-metadata/README.md
```diff
@@ -111,4 +111,37 @@ module.exports = function (defaults) {
...
+```js
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function (defaults) {
+  function customNormallizedFilePath(filePath, packageName, projectRoot) {
+    // Your custom normalize process
```
drewlee, line 124: Nit:
```
// Custom normalization
```

packages/babel-plugin-ember-test-metadata/__tests__/get-normalized-file-path-test.js
drewlee: Should be something along the lines of:
```js
function normalizeBuildIsolationPath({ filePath, packageName }) {
  const pathSegments = filePath.split(sep);
  const relativePathSegments = pathSegments.slice(pathSegments.indexOf(packageName) + 1);
  const normalizedPathSegments = relativePathSegments.filter((segment) => segment !== packageName);

  return normalizedPathSegments.join(sep);
}
```

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
drewlee, line 23: Would also pass the arguments through as an object, so that the custom function can reference only the parameters it needs.

```js
return getCustomNormalizedFilePath({ packageName, isUsingEmbroider, projectRoot, filename, root });
```

drewlee (approved):

PR #235, opened 2023-02-18
Update `getCustomNormalizedFilePath` from a function to a resolved path

zhanwang626:
## Summary
Previously the `getCustomNormalizedFilePath` babel option is a function, this results a problem in the build process.
```shell
strict mode: "items" is 1-tuple, but minItems or maxItems/additionalItems are not specified or different at path "#/properties/views"
[broccoli-persistent-filter:Babel > [Babel: appreciation-test-helpers]: Babel: appreciation-test-helpers] was configured to `throwUnlessParallelizable` and was unable to parallelize a plugin.
```
Instead we can change the `getCustomNormalizedFilePath` option's type to a filePath, so we can walk around the problem.

### DIFF
--- a/packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
+++ b/packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
@@ -18,10 +18,11 @@ const {
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
 function getNormalizedFilePath({ packageName, getCustomNormalizedFilePath, isUsingEmbroider, projectRoot, filename, root }) {
-  if (typeof getCustomNormalizedFilePath === 'function') {
+  if (getCustomNormalizedFilePath) {
+    const customNormalizedFilePath = require(getCustomNormalizedFilePath);
     const options = { packageName, isUsingEmbroider, projectRoot, filename, root };
   
-    return getCustomNormalizedFilePath(options);
+    return customNormalizedFilePath(options);
   } else if (!isUsingEmbroider) {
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
```diff
@@ -18,10 +18,11 @@ const {
  * @returns {string} E.g. tests/acceptance/my-test.js
  */
 function getNormalizedFilePath({ packageName, getCustomNormalizedFilePath, isUsingEmbroider, projectRoot, filename, root }) {
-  if (typeof getCustomNormalizedFilePath === 'function') {
+  if (getCustomNormalizedFilePath) {
+    const customNormalizedFilePath = require(getCustomNormalizedFilePath);
     const options = { packageName, isUsingEmbroider, projectRoot, filename, root };
   
-    return getCustomNormalizedFilePath(options);
+    return customNormalizedFilePath(options);
```
scalvert, line 25: Should we verify this is a path that exists? Also, if it does exist, should we cache it for the remainder of the build?

For instance, we can do this check once, and if it's a valid path we can require the function, and cache it locally for the duration of the build.

scalvert (changes requested):

packages/babel-plugin-ember-test-metadata/get-normalized-file-path.js
scalvert: Actually cancel this; node will already cache the require and returned the reified module if called again within the same process.

scalvert (approved):

PR #241, opened 2023-05-16
[Bug] Fix test file paths for Ember addon project

zhanwang626:
Currently `babel-plugin-ember-test-metadata` can not report correct path for Ember addon. For example, it will report filePath like 

`"filePath":"dummy/tests/integration/components/metadata/skills-test.js" `

instead of

`"filePath":"tests/integration/components/metadata/skills-test.js" `

This PR will fix this problem by removing "dummy" in the path segments

### DIFF
--- a/packages/babel-plugin-ember-test-metadata/get-relative-paths.js
+++ b/packages/babel-plugin-ember-test-metadata/get-relative-paths.js
@@ -14,7 +14,7 @@ function _getRelativeProjectPath(pathSegments, projectDir, projectRoot) {
 
 function _getRelativePathForClassic(filePath, packageName, projectRoot) {
   const projectDir = _getNormalizedPackageDir(packageName);
-  const pathSegments = filePath.split(sep).filter(item => item !== 'dummy')
+  const pathSegments = filePath.split(sep).filter(segment => segment !== 'dummy')
   const testFilePath = pathSegments
     .splice(pathSegments.lastIndexOf(projectDir) + 1)
     .join(sep);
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-ember-test-metadata/get-relative-paths.js
```diff
@@ -14,7 +14,7 @@ function _getRelativeProjectPath(pathSegments, projectDir, projectRoot) {
 
 function _getRelativePathForClassic(filePath, packageName, projectRoot) {
   const projectDir = _getNormalizedPackageDir(packageName);
-  const pathSegments = filePath.split(sep);
+  const pathSegments = filePath.split(sep).filter(item => item !== 'dummy')
```
drewlee, line 17: Minor nitpick for formatting and variable naming:
```suggestion
  const pathSegments = filePath.split(sep).filter((segment) => segment !== 'dummy')
```

packages/babel-plugin-ember-test-metadata/get-relative-paths.js
```diff
@@ -14,7 +14,7 @@ function _getRelativeProjectPath(pathSegments, projectDir, projectRoot) {
 
 function _getRelativePathForClassic(filePath, packageName, projectRoot) {
   const projectDir = _getNormalizedPackageDir(packageName);
-  const pathSegments = filePath.split(sep);
+  const pathSegments = filePath.split(sep).filter(segment => segment !== 'dummy')
```
drewlee, line 17: Missing ending semicolon. Also, let's wrap the `segment` param in parentheses to be consistent with the rest of the code base. Not sure why these aren't getting caught by ESLint.

drewlee (approved):
Thanks!

